PR #281, opened 2026-07-18
feat(ci): bounded retry-with-backoff for Unity serial activation (build-lock #57)

wallstop:
## What

Propagates the proven unity-helpers change (PR #306) into DxMessaging: `Invoke-UnityLicenseActivate` in `scripts/unity/run-ci-tests.ps1` now retries transient Unity serial-activation contention within a bounded wall-clock budget with jittered exponential backoff, instead of failing the whole run on the first `20111`.

## Why (build-lock #57)

The organization build lock's release cooldown is deliberately near-zero. When this run acquires a Unity seat, the **prior** holder's activation may not have propagated as *returned* yet, so Unity answers error **20111 "maximum number of activations."** That is transient seat contention, not a hard error — retrying briefly lets the seat free instead of the run failing (or the lock blindly holding a slot for minutes).

## The 20111 retry-but-preserve-incident design

The subtlety is that `20111` is **two different things**:

- a **transient handoff** (previous seat still freeing) that clears within the budget, or
- a **real capacity leak** that persists past the budget.

So the change retries `20111` **only within a bounded budget** and, crucially, treats a `20111` that *survives* the deadline as a genuine account incident:

- **Log semantics are OVERWRITE-per-attempt** (`Tee-Object -FilePath`, no `-Append`). A success after a transient `20111` leaves a **clean** final log, so the existing return-side classifier never mis-reads a stale `20111` as an incident. A **persistent** `20111` leaves the `20111` intact in the final-attempt log, so the existing return-side account-incident classifier still fires. We never treat `20111` as success and never reset the incident path.
- **Permanent failures fail fast**: serial expiry (`20113`) throws immediately with no retry.
- **Every throw** names the reason + attempt count + the non-uploaded activation log path — **never** the serial/email/password.

### Shape

Three pure, StrictMode-safe helpers added before the function:

- `Get-UnityActivationRetryBudgetSeconds` — honors `DXM_ACTIVATION_RETRY_BUDGET_SECONDS` (DxMessaging's `DXM_` prefix), default `360s`. Non-integer/negative → `::warning::` + default; **`0` = explicit opt-out** (single attempt, legacy fail-fast). Note `0` differs from the `*-TIMEOUT_SECONDS` helpers where `0` means "unbounded", because an unbounded activation retry could pin a lock slot forever on a real capacity leak.
- `Get-UnityActivationRetryDelaySeconds` — `min(cap, base·2^(N-1))` backoff ceiling (base 5s, cap 30s), overflow-clamped.
- `Get-UnityActivationFailureClass` — `success`/`retryable`/`hard` classifier. The `20111`/`20113` matchers use non-digit boundary guards (`(?<![0-9])20111(?![0-9])`) so `201110` / `120111` do **not** match.

The refactored `Invoke-UnityLicenseActivate` gains a production-inert `-ActivationInvoker` testability seam (never passed in production; the real `& $EditorPath … | Tee-Object` path is byte-for-byte unchanged when it is `$null`) and a `-RetryBudgetSeconds` param.

## Tests

`scripts/__tests__/test-unity-license-activation-retry.ps1` (mirrors the reference's pwsh test) extracts the real function definitions from `run-ci-tests.ps1` via the PowerShell AST and exercises them **without launching Unity**:

- classifier table, **including the `201110`/`120111` boundary-guard rows** that must NOT match `20111`;
- backoff ceiling table;
- `DXM_ACTIVATION_RETRY_BUDGET_SECONDS` parsing (default / override / opt-out / invalid);
- retry loop via the fake invoker: first-success; `20111`-then-success asserts a **clean** final log; persistent-`20111` asserts throw + preserved `20111` + no credential leak; hard-fail-fast; budget-`0` single attempt.

Wired into the `script-tests` job in `ci.yml` as a `shell: pwsh` step (runs on the ubuntu/macos/windows matrix). A JS test was impractical here — the repo's JS LoC budget is at its `13700` ceiling — so this matches the reference's pwsh route.

Local run: **62 passed, 0 failed.** `node --test scripts/` matches the clean-master control exactly (no new failures).

---

> [!NOTE]
> **Medium Risk**
> Changes CI license activation behavior and credential-adjacent logging paths; mitigated by bounded retries, fail-fast on expiry, tests, and preserved incident semantics for persistent 20111.
> 
> **Overview**
> **`Invoke-UnityLicenseActivate`** in `run-ci-tests.ps1` no longer fails on the first Unity **20111** (“maximum activations”) when the org build lock hands off a seat. It retries within a **bounded wall-clock budget** (default 360s via `DXM_ACTIVATION_RETRY_BUDGET_SECONDS`, `0` = single attempt) with **jittered exponential backoff**, classifies outcomes (`20111` retry, `20113` fail-fast, others retry within budget), and **overwrites the activation log each attempt** so a later success clears stale `20111` while a budget-exhausted `20111` still leaves evidence for the existing return-side incident path. Credentials stay out of error text.
> 
> New **pure helpers** (`Get-UnityActivationRetryBudgetSeconds`, `Get-UnityActivationRetryDelaySeconds`, `Get-UnityActivationFailureClass`) support testing; production adds an optional **`-ActivationInvoker`** seam only for tests.
> 
> **`scripts/__tests__/test-unity-license-activation-retry.ps1`** AST-loads those functions and runs table-driven tests (classifier boundaries, backoff, env parsing, fake retry loop). **CI** runs it in the `script-tests` matrix via a new `pwsh` step.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1ddacac0b45acfbf298af86f12628239e352dc41. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -724,6 +724,11 @@ jobs:
         shell: bash
         run: node --test scripts/
 
+      - name: Test Unity license activation retry
+        if: ${{ needs.changes.outputs.scripts != 'false' }}
+        shell: pwsh
+        run: ./scripts/__tests__/test-unity-license-activation-retry.ps1 -VerboseOutput
+
       - name: Run validators
         if: ${{ needs.changes.outputs.scripts != 'false' && matrix.os == 'ubuntu-latest' }}
         shell: bash
--- a/scripts/__tests__/test-unity-license-activation-retry.ps1
+++ b/scripts/__tests__/test-unity-license-activation-retry.ps1
@@ -1,0 +1,267 @@
+#!/usr/bin/env pwsh
+<#
+.SYNOPSIS
+    Data-driven tests for the Unity serial-activation retry policy in
+    scripts/unity/run-ci-tests.ps1 (build-lock issue #57).
+
+.DESCRIPTION
+    The organization build lock's release cooldown is deliberately near-zero, so a
+    freshly acquired seat can hit Unity error 20111 ("maximum number of
+    activations") while the previous holder's seat is still propagating as
+    returned. Invoke-UnityLicenseActivate now retries that transient contention
+    within a bounded wall-clock budget with jittered exponential backoff, while
+    failing fast on permanent errors and preserving the 20111 evidence for the
+    account-incident path when contention does NOT clear.
+
+    This runner extracts the real function definitions from run-ci-tests.ps1 via
+    the PowerShell AST (the script's top-level param()/execution make it
+    non-dot-sourceable) and exercises them WITHOUT launching Unity:
+      * the pure classifier / backoff / budget helpers via data-driven tables, and
+      * the full retry loop via Invoke-UnityLicenseActivate's testability seam
+        (-ActivationInvoker), a fake editor that is deterministic and needs no
+        external process, so this runs on ubuntu-latest, macos-latest, and Windows
+        alike.
+#>
+[CmdletBinding()]
+param(
+    [switch]$VerboseOutput
+)
+
+Set-StrictMode -Version Latest
+$ErrorActionPreference = 'Stop'
+
+$scriptRoot = Split-Path -Parent $PSScriptRoot
+$target = Join-Path $scriptRoot 'unity/run-ci-tests.ps1'
+if (-not (Test-Path -LiteralPath $target -PathType Leaf)) {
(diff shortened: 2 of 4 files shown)


wallstop:
bugbot run

PR #284, opened 2026-07-19
ci: classify Unity cleanup without Python

wallstop:
## What changed

- recognize Unity's exact current entitlement-return plus legacy-ULF-skip evidence
- keep the existing two-signal, fail-closed cleanup proof contract
- port bounded account-health evidence classification from Python to the already-required PowerShell runtime
- fail composite outputs closed when health classification fails or returns malformed/missing output
- preserve bounded `.log`/`.txt` scanning, reparse-point avoidance, exact 20111 token boundaries, sanitized diagnostics, and SHA-256 evidence digests

## RCA

Default-branch run `29698538543` was intentionally cancelled after a perf-doc commit superseded its head. Its active job returned the Unity entitlement successfully, but the cleanup classifier did not recognize Unity's current exact line:

`[Licensing::Module] Error: Serial number unavailable for ULF return; skipping operation`

The secondary account-health step then failed because `DAD-MACHINE` has no usable Python runtime. The lock correctly quarantined capacity instead of assuming safety. Exact same-runner logs proved the entitlement return, so reservation `926c489d-2984-46de-8b72-456422911bd7` was recovered through guarded central run `29699406532` with `resource-safe=true`.

## Red / green

- red: canonical captured return wording failed the prior classifier
- red: workflow contract rejected the undeclared Python runtime dependency
- green: both PowerShell classifier self-tests pass
- green: focused workflow lifecycle contract passes
- green: full `validate:all` with pinned .NET 9.0.314, pre-commit, Prettier/yamllint, and diff hygiene
- full Node suite: 245/247 locally; the two known Windows-host baseline failures require symlink privilege and WSL path conversion and are covered on Linux CI
- two adversarial reviews: zero remaining issues

Follow-up to #283 and Ambiguous-Interactive/ambiguous-organization-build-lock#47.

---

> [!NOTE]
> **Medium Risk**
> Changes affect organization build-lock admission signals on self-hosted Unity runners; misclassification could quarantine capacity or admit unsafe cleanup, though behavior is fail-closed with expanded self-tests.
> 
> **Overview**
> **Unity license return** now treats Unity’s current module-prefixed log lines as valid cleanup proof: entitlement success plus either the full ULF-skip error line or legacy bare / explicit ULF-return wording, still requiring both signals and rejecting near-misses and abnormal exit codes (including additional Windows termination codes).
> 
> **Account-health classification** moves from `classify_unity_account_health.py` into `Classify-UnityAccountHealth.ps1`, so the composite action no longer needs Python on the runner while keeping bounded `.log`/`.txt` scanning, reparse-point avoidance, exact `20111` token matching, and SHA-256 evidence digests.
> 
> **Composite outputs** for `resource-safe` and `resource-cleanup-status` are tightened so they are only true/`confirmed` when the health step succeeds with `healthy` *and* the return step’s outputs match; otherwise they fail closed to `false`/`unknown`. Validation and CI contract tests were updated to run the PowerShell self-tests and assert the new wiring.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7794e54f238d3ad14578144b64e0cd1ab6d178b4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


wallstop:
@cursor review
@copilot review

Please review exact head `7794e54f238d3ad14578144b64e0cd1ab6d178b4`, focusing on exact two-signal cleanup proof, Python-to-PowerShell classifier parity, fail-closed composite outputs, bounded/symlink-safe evidence traversal, sanitized diagnostics, and cross-platform path semantics.

PR #293, opened 2026-07-29
feat: adopt the Agent Skills standard for .llm and a tested Unity MCP CLI

wallstop:
## Summary

Two modernizations, using the private `Ambiguous-Interactive/qora-redux` repository as the reference implementation, and improving on it in both cases.

**1. Unity MCP** — `scripts/mcp/unity-mcp.mjs` replaces 404 lines of untested shell/PowerShell and the runtime `supergateway` dependency with one tested ESM module exposing `probe`, `configure`, and `bridge`.

**2. `.llm` harness → [Agent Skills](https://agentskills.io/specification) standard** — 159 flat documents become 30 skill directories that Claude Code, Codex, and GitHub Copilot discover natively. `.llm/` stays the single source of truth.

## Beyond the qora reference

| | qora | here |
| --- | --- | --- |
| Endpoint discovery | one fixed host/port | walks candidate hosts (`host.docker.internal`, `127.0.0.1`, `/etc/resolv.conf` nameservers, `/proc/net/route` gateways) x ports (`9020`, `9003`) until an MCP `initialize` handshake succeeds |
| Probe failures | pass/fail | classified `unreachable` / `unauthorized` / `http-error` / `malformed`, because the fix differs |
| Repo root vs Unity project | conflated (qora *is* a Unity project) | separated. DxMessaging is a UPM package whose host project lives elsewhere, so `--project` is validated for `bridge` alone and `probe`/`configure` work inside the container. qora's version would fail there. |
| Dependencies | `express` | `node:http`, keeping a published package's dependency surface smaller |
| Skill frontmatter | `name: Manage Vendor Pointer Files`, `allowedTools` | spec-conformant: `name` matches the directory, `allowed-tools`, repo-specific fields under `metadata` |

An explicit `--host`/`--port` is always probed first, so discovery can never override a deliberate setting.

## Data behind the skills migration

Measured before deciding, not guessed:

- **8,484 of 28,518 lines (29%)** of the old harness was YAML frontmatter, averaging 53 lines per file of `complexity.reasoning` and `impact.*.details` that no agent reads. Stripping it dropped files over 200 lines from **63 to 14**.
- The 159→30 mapping was verified as a **bijection** before anything moved: no duplicate assignment, no unmapped file, every name spec-valid.
- A heading-level audit across all 159 migrated files confirms **zero content loss**.
[...]

PR #294, opened 2026-07-29
fix(mcp): make probe --no-discover actually probe

wallstop:
Fixes the one issue [Cursor Bugbot reported on #293](https://github.com/Ambiguous-Interactive/DxMessaging/pull/293#discussion_r2604060000).

## The bug

`resolveEndpoint` returned early with no `found` and an empty `attempts` list whenever discovery was off, so `runProbe`'s `if (!found)` always fired:

```
$ node scripts/mcp/unity-mcp.mjs probe --no-discover --host 127.0.0.1 --port 9020
unity-mcp: No Unity MCP endpoint responded. Attempts:
                                                        <- nothing
```

`probe --no-discover` was unusable, and the failure said nothing about why.

## The fix

`--no-discover` meant "skip the handshake", which makes `probe` pointless: verifying an endpoint is the command's entire job. It now means **narrow the candidate list to the configured host and port**. The single endpoint is still probed, and a failure names it:

```
unity-mcp: No Unity MCP endpoint responded. Attempts:
  http://127.0.0.1:9020/mcp - unreachable (no TCP listener)
```

Against the live host bridge it now correctly reports `unauthorized (HTTP 401)` instead of an empty list.

`configure` is unchanged in behaviour: with discovery off it writes the configured endpoint whether or not the probe answers, and still refuses when the probe comes back `unauthorized`.

## Why the tests missed it

Every existing `runProbe` / `runConfigure` test injected `runtime.candidates`, which bypasses exactly the narrowing that was broken. The three new tests deliberately pass **no** candidates, so the endpoint has to come from `options.host` / `options.port`. One asserts the failure message names the endpoint rather than being empty.

`node --test scripts/` 403 pass / 0 fail. Docs updated in `scripts/mcp/README.md`, the `unity-mcp-test-loop` skill, and `--help`.

## Note

Copilot declined to review #293 three times because it exceeded its 300-file limit, so Bugbot was the only automated reviewer that ran. This PR is 4 files.

---

> [!NOTE]
> **Low Risk**
> Localized fix to Unity MCP CLI discovery/probe behavior plus docs and tests; no auth, bridge, or transactional config logic changes beyond clearer failure reporting when `--no-discover` is used.
> 
> **Overview**
> **`--no-discover` now limits probing to the configured host/port instead of skipping the MCP handshake**, fixing `probe --no-discover`, which always failed with an empty "Attempts" list.
> 
> `resolveEndpoint` routes through `discoverEndpoint` with a single candidate when discovery is off, keeps the configured endpoint when nothing answers (so **`configure` still writes the named URL**), and **`probe` failures include classified attempt details** (e.g. unreachable or unauthorized). The no-handshake warning for configure applies whenever the probe fails, not only with full discovery.
> 
> Docs (`scripts/mcp/README.md`, unity-mcp skill, `--help`) and a regression test without injected `candidates` align with the new semantics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 37b06d4303217ea1e21b6f9033a3b1923877c022. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


wallstop:
bugbot run

PR #291, opened 2026-07-27
ci: centralize Unity cleanup policy

wallstop:
## Summary

- delegate Unity cleanup evidence classification to the central organization policy pinned at `673eb65e7d863a1a8a8a70882bd980e189d41754`
- release all six licensed lock windows through the matching central action and require the typed final cleanup gate
- keep raw Unity return evidence private, delete it after classification, and remove the duplicated local classifiers
- update direct development dependencies to the latest available versions

## Why

This completes the remaining DxMessaging rollout for Ambiguous-Interactive/ambiguous-organization-build-lock#85. Central ownership prevents cleanup semantics from drifting between repositories, while preserving the existing native PowerShell Unity execution and matrix shape.

The previous return pipeline also copied Unity account/serial-bearing output into public workflow logs. The replacement writes it only to private runner-temp evidence consumed by the central classifier.

## Impact

No Unity matrix rows, licensed jobs, lock capacity, or organization settings change. Existing jobs gain fail-closed cleanup classification and final release coherence checks.

## Validation

- `npm test` — 238 passed, 9 skipped because PowerShell is unavailable, 0 failed
- focused CI workflow contract — 13 passed
- formatting, Markdown, spelling, JS LOC budget, package/meta, Unity version and PR-policy validators
- `npm outdated --json` — no outdated direct dependencies
- temporary lockfile `npm audit --audit-level=low` — 0 vulnerabilities
- actionlint across all workflows
- `git diff --check`

## Review record

Main-thread adversarial fallback was used because this session does not authorize sub-agent delegation. Round 1 found unused local classifier copies that would preserve policy drift; they and their obsolete validation command were removed. Fresh validation passed, and round 2 found no actionable findings.

---

> [!NOTE]
> **Medium Risk**
> Changes org-wide Unity seat release semantics and can fail jobs when cleanup cannot be confirmed; mitigated by pinning central actions and conservative private logging, but misclassification or gate regressions could block CI or delay lock recovery.
> 
> **Overview**
> Unity license return and org-lock teardown no longer use **local PowerShell classifiers**; cleanup proof and account health are decided by the pinned **ambiguous-organization-build-lock** actions (`classify-unity-cleanup-evidence` / `require-confirmed-unity-cleanup` at `673eb65`).
> 
[...]

PR #278, opened 2026-07-16
Fail closed when Unity runners are unavailable

wallstop:
## Summary
- replace five PAT/GITHUB_TOKEN soft-pass probes with the v1.8.2 reader-App preflight
- pin all six Unity lock lifecycles to the schema-5-capable v1.8.2 SHA
- preserve same-repository PR Unity validation without environments or manual approvals
- keep every aggregate verdict fail closed when preflight or Unity work fails

## Verification
- focused CI contract: 13/13 pass
- validate:all policy, version, line-budget, and license-classifier gates pass (local analyzer payload remains limited by the repo-pinned .NET 9.0.314 SDK not installed on this host)
- actionlint built-in validation passes for all five workflows
- Prettier check passes
- full npm test baseline-only Windows/WSL portability failures reproduced on untouched master

---

> [!NOTE]
> **Medium Risk**
> Unity CI and release gates now depend on org reader App secrets and will hard-fail if misconfigured, whereas the old soft-pass could still queue indefinitely; lock action upgrades touch every licensed job path.
> 
> **Overview**
> Replaces duplicated inline `gh api` runner probes in five Unity workflows with the shared **`check-unity-runner-availability`** action at **v1.8.3**, wired to **`BUILD_LOCK_READER_APP_*`** secrets and per-workflow **`required-label-sets`** (perf keeps the extra **`fast`** label).
> 
> **Preflight behavior shifts from soft-pass** (warn and continue when `GITHUB_TOKEN`/`RUNNER_AUDIT_PAT` could not list runners) **to fail-closed**: missing reader credentials, unreadable inventory, ACL gaps, or no matching online runner fails before self-hosted jobs queue. Job-level **`actions: read`** on preflight is dropped where it only served the old probe.
> 
> All **`acquire-build-lock`** / **`release-build-lock`** pins move from **v1.7.1** to the same **v1.8.3** SHA across perf, release, benchmarks, GameCI experiment, and unity-tests. Docs and **`ci-aggregate-workflow.test.js`** are updated to match the new contract and drop **`RUNNER_AUDIT_PAT`** guidance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0ce87d7ab9a2c98ebab82bbaf36088326c5fb656. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #295, opened 2026-07-29
ci: unblock Dependabot PRs; feat(editor): Message subscriptions inspector section

wallstop:
## Summary

Two things, both blocking work that was already queued.

1. **Every Dependabot pull request has been permanently red since 2026-07-17.** Both causes are fixed, and #292 is incorporated here with the pin split it would have introduced repaired.
1. **PLAN.md M2 ships:** a themed **Message subscriptions** section on every `MessageAwareComponent` inspector, which also adopts the `.dx-inspector` / `.dx-sub*` stylesheet surface that until now was defined in USS and referenced by no C#.

Closes #292.

### Dependabot jobs cannot read the organization secrets

#278 added a `runner-preflight` job calling `check-unity-runner-availability` with `secrets.BUILD_LOCK_READER_APP_ID`. Dependabot runs read from the separate **Dependabot** secret store (`Secret source: Dependabot` in the job log), so that secret resolves empty and the fail-closed action aborts before it checks anything:

```text
##[error]Unity runner preflight failed closed: Missing required input: reader-app-id.
```

`Unity CI Success` needs `runner-preflight`, and branch protection requires `Unity CI Success`. The licensed matrix could not have passed either -- `UNITY_SERIAL` is empty there for the same reason.

`runner-preflight` and the `unity-tests` matrix now exclude Dependabot pull requests exactly the way they already exclude forks, and the always-reporting `unity-ci-success` gate treats that as a third intentional skip shape. Dependency bumps stay validated by the full matrix on the post-merge push to master.

### Tests hardcoded the SHA that Dependabot exists to change

`scripts/__tests__/ci-aggregate-workflow.test.js` and `scripts/validate-unity-pr-policy.py` pinned the build-lock action SHA as a literal, so any bump of that action failed Script tests on all three runners plus the devcontainer smoke job that runs `npm test`. Both now derive the pin from the tree and assert what actually matters: each action group is immutably pinned and identical at every call site.

### The pin split

#292 bumped the workflow acquire/preflight refs to v1.10.0, moved release/require-confirmed to an untagged default-branch commit, and left `classify-unity-cleanup-evidence` alone -- it lives in `.github/actions/return-unity-license/action.yml`, which Dependabot's `github-actions` ecosystem does not scan under `directory: "/"`. That would have split the centralized Unity cleanup policy from #291 across two SHAs.

All six build-lock references are unified on `3741b56 # v1.10.0`; v1.10.0 has `673eb65` (the cleanup-policy commit) as an ancestor, so one pin covers the group. Its manifest changes are additive: two new `release-build-lock` outputs, and a preflight matching registered rather than only online runners. `.github/dependabot.yml` now lists the composite-action directory so the pin moves with the group.

## Part 2: the Message subscriptions section

Under the default inspector body, every `MessageAwareComponent` now lists the registrations its live `MessageRegistrationToken` holds: message type, registration kind, priority, observed call count, and a live/idle dot.

- **The dot means "subscribed on the bus," not "has traffic."** A token disabled by `OnDisable` is the most common reason a handler stops firing, and that is what it reports.
- **Unknown is not zero.** With diagnostics off a row reads `calls n/a`; `0 calls` would assert a handler never fired when nothing was recording.
- Reading the lazy `_callCounts` dictionary is gated on the token actually recording, so an open inspector never materializes it on a token whose owner never enabled diagnostics.
- Registrations change from game code with no editor event to hook, so the section polls at 500 ms while on screen and rebuilds only when a cheap revision signal moves.
- Hidden under multi-object selection, because the rows describe one component.
- Per PLAN.md DEC-5, capture and formatting live in a plain-C# state object, so most of the ten new tests need no editor panel and run version-free across the 2021.3+ matrix.

## Verification

- Unity MCP EditMode loop (Unity 6000.4.6f1): **468 passed / 0 failed** in 14.3 s.
- `npm test`: 400/400. `npm run validate:all`: clean, including the JS LOC budget.
- Both new guards proven red-green before committing: removing the Dependabot exclusion fails `validate-unity-pr-policy.py`, and splitting the cleanup-policy pin fails `ci-aggregate-workflow.test.js` with the offending files named.

## Needs a real Dependabot run to confirm

Whether the `directories` form is honored for the `github-actions` ecosystem. If it is not, the new consistency guard fails closed with a message naming the exact files, which is the safe direction.

## Superseded work dropped

`dev/wallstop/even-better-ui` (one commit, 2025-06-17, 252 commits behind) threaded an `ImmutableArray<InstanceId> relevantReceivers` through `MessageEmissionData` to join an emission to the handlers that observed it. That already shipped allocation-free as the `traceId` + `registrationHandle` pair in #261; the branch also allocated per emission and depended on `System.Collections.Immutable`, which is not part of the Unity 2021.3 floor. Recorded in PLAN.md so it is not rediscovered.

---

> [!NOTE]
> **Medium Risk**
> Changes to Unity CI gating and organization lock/cleanup pins affect required checks and licensed runner behavior; the editor work is additive but reads token internals for diagnostics display.
> 
> **Overview**
> Unblocks **Dependabot** the same way forks are handled: `runner-preflight` and the licensed `unity-tests` matrix skip when `dependabot[bot]` opens the PR (org Unity and build-lock secrets are not in the Dependabot store), and **Unity CI Success** treats that as a third intentional skip alongside fork and docs-only PRs. Policy tests no longer hardcode build-lock SHAs—they derive pins from the tree and assert one immutable SHA per action group.
> 
> **CI hygiene:** all `ambiguous-organization-build-lock` references move to **`3741b56` (# v1.10.0)** so acquire/preflight and release/cleanup/classify stay on one pin; **Dependabot** also scans `/.github/actions/return-unity-license` so composite `classify-unity-cleanup-evidence` bumps with workflows. Routine bumps to `actions/checkout`, `setup-node` v7, `setup-dotnet` v6, `setup-python` v7, and related actions land across workflows.
> 
> **Editor (PLAN M2):** `MessageAwareComponent` fallback inspectors gain a **Message subscriptions** UI Toolkit section (`.dx-inspector` / `.dx-sub*`) listing live `MessageRegistrationToken` rows—type, registration kind, priority, call count or `calls n/a`, and live/idle dot—with 500 ms polling and revision-based rebuilds; hidden for multi-select. Changelog and `docs/guides/inspector-overlay.md` document it; EditMode tests cover capture and rendering.
> 
> Also: `.gitignore` adds `GOAL.md`, and the Unity MCP test-loop doc warns about stale assemblies when compile fails.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3d0da6d0d0f4e0a4ea1dd6c57d7caa63719c506b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/actions/return-unity-license/action.yml
+++ b/.github/actions/return-unity-license/action.yml
@@ -105,7 +105,7 @@ runs:
     - name: Classify prior exact return evidence
       id: classify_prior
       if: ${{ steps.prior_metadata.outputs.prior-ready == 'true' }}
-      uses: Ambiguous-Interactive/ambiguous-organization-build-lock/.github/actions/classify-unity-cleanup-evidence@673eb65e7d863a1a8a8a70882bd980e189d41754
+      uses: Ambiguous-Interactive/ambiguous-organization-build-lock/.github/actions/classify-unity-cleanup-evidence@3741b56ceab4a68ba4c09fe7e91e804b53ff2412 # v1.10.0
       with:
         return-log-path: ${{ steps.prior_metadata.outputs.return-log-path }}
         return-command-completed: "true"
@@ -174,7 +174,7 @@ runs:
     - name: Classify redundant return evidence
       id: classify_return
       if: ${{ always() && steps.classify_prior.outputs.resource-safe != 'true' }}
-      uses: Ambiguous-Interactive/ambiguous-organization-build-lock/.github/actions/classify-unity-cleanup-evidence@673eb65e7d863a1a8a8a70882bd980e189d41754
+      uses: Ambiguous-Interactive/ambiguous-organization-build-lock/.github/actions/classify-unity-cleanup-evidence@3741b56ceab4a68ba4c09fe7e91e804b53ff2412 # v1.10.0
       with:
         return-log-path: ${{ steps.capture_return.outputs.return-log-path }}
         return-command-completed: ${{ steps.capture_return.outputs.return-command-completed }}
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,17 @@
 version: 2
 updates:
-  # GitHub Actions workflow updates
+  # GitHub Actions workflow updates.
+  #
+  # A bare "/" makes Dependabot scan .github/workflows plus a repository-root
+  # action.yml, and nothing else. The build-lock pin inside
+  # .github/actions/return-unity-license/action.yml therefore sat outside every
+  # bump, so a bump split the centralized Unity cleanup policy across two SHAs.
+  # Listing that directory keeps the whole group on one pin; the split now fails
+  # scripts/__tests__/ci-aggregate-workflow.test.js closed.
   - package-ecosystem: "github-actions"
-    directory: "/"
+    directories:
+      - "/"
+      - "/.github/actions/return-unity-license"
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
       yaml: ${{ steps.filter.outputs.yaml }}
     steps:
       - name: Checkout
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           fetch-depth: 1
           persist-credentials: false
@@ -205,7 +205,7 @@ jobs:
 
       - name: Checkout
         if: ${{ needs.changes.outputs.actionlint != 'false' }}
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           persist-credentials: false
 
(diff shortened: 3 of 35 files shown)


wallstop:
bugbot run

wallstop:
bugbot run

wallstop:
Reviewer follow-up, both Bugbot findings fixed:

- **Stale subscription rows after swap** -- `Revision` now folds every rendered field, including each row's message type, registration kind, priority, and live state. Regression test swaps one registration for another with diagnostics off (so every call count is `UnknownCallCount`) and asserts the revision moves. Fixed in 233b2b8d.
- **Subscription rows lack content padding** -- the rows body now carries the same 11px horizontal inset as `.dx-inspector__head`, so names clear the border and the trailing status dot cannot clip on the corner radius. The inset lives in C# rather than on `.dx-sub` so the migrated stylesheet stays identical to the design-system spec; the test asserts a non-zero inset rather than pinning a pixel value. Fixed in 3d0da6d0.

Bugbot's latest pass reports no remaining issues.

Two bugs CI caught that the local MCP loop could not: `CS0103` on `UnknownCallCount` (unqualified cross-type reference) and `CS0117` from `RefreshInspectorWarning` being removed alongside an unused seam. The host editor was serving a stale assembly with Auto Refresh disabled, so a failed compile reported the previous run's passing numbers. `.llm/skills/unity-mcp-test-loop` now carries a freshness check for that.

Final state: 34 checks pass, 7 intentional skips, all nine Unity legs green across 2021.3 / 2022.3 / 6000.3. Local MCP EditMode 470/470 on a verified-fresh assembly.

GitHub Copilot returned "reached their quota limit" on all five review requests, so it has no verdict on this PR.

PR #302, opened 2026-07-29
feat(editor): live Message Monitor mode

wallstop:
## Summary

Ships PLAN.md **M3**: a **Live** mode beside the Message Monitor's snapshot view (WS-5 / DEC-1), plus the **WS-2.2** adoption of the columnar recorder half of the editor stylesheet.

Those classes -- `.dx-record`, `.dx-chip`/`.dx-filter`, `.dx-list-header`/`.dx-col-*`, `.dx-row`/`.dx-row--alt`/`.dx-row__*`, `.dx-detail*`, `.dx-kv*` and `.dx-footer*` -- shipped with the theme but no C# referenced them, so the stylesheet described a surface the windows never rendered. They are all rendered now, and a contract test pins the whole set so dropping one fails CI rather than silently returning it to dead style.

## Design

Two deliberate departures from the Appendix C sketch in PLAN.md, both forced by the plan's own do-not-violate list:

**Pull, not push.** Appendix C proposed a static `ConcurrentQueue` ingest tap. That needs a runtime hook, which violates "editor-only, zero runtime cost". The bus already rings every emission while diagnostics are on, and every record carries a monotonic dispatch sequence number (`MessageEmissionData.traceId`), so `MessageMonitorLiveRecorder` polls that ring every 250 ms and folds in whatever sits past its cursor. **Nothing under `Runtime/` changed.** Two consequences worth calling out:

- A jump in the sequence means the bus ring overwrote emissions before the drain reached them. The gap is exact, so it is reported as `missed` in the footer instead of the log quietly having a hole.
- The window skips capturing a snapshot at all while `MessageBus.EmissionId` still equals the cursor, so an idle scene costs one integer compare per poll rather than rebuilding an entry per buffered record.

**Coalescing.** An adjacent run of identical emissions (same message type, context and route kind) folds into one row with a count. A bus can emit the same message thousands of times per second, so this is what keeps the log readable, what bounds the rendered row count, and what gives `.dx-col-count` / `.dx-row__count` a real meaning rather than a force-fit one.

## Structure

- `Editor/Windows/MessageMonitorLiveRecorder.cs` -- the bounded log. No GUI or editor type, clock injected, so draining, coalescing, pausing, bus-reset rebasing and loss accounting are all covered by plain NUnit with no panel and no version gate (per DEC-5).
- `Editor/Windows/DxMessagingMessageMonitorLiveView.cs` -- the columnar UITK surface. The toolbar is built once and the body re-rendered separately, so typing in the filter does not rebuild and unfocus the field being typed into.
- `DxMessagingMessageMonitorWindow` gains a Live button, a serialized mode flag, the poll pump, and a play-mode hook that clears the log (play-mode transitions reset the bus and restart its dispatch sequence).

The snapshot view is untouched; live mode is purely additive.

## Bugs the red phase caught

1. `default(MessageMonitorLiveViewState)` hid every route kind. The chip flags were stored as `ShowX`, so the struct's default -- reachable from an omitted optional parameter, an uninitialized field, or `Default` itself -- meant "hide everything" and rendered an empty log. Now stored inverted so `default` is an unfiltered log.
2. Bus-reset detection ran after the collect, so `Ingest` returned early on an empty pending list and a reset bus was ignored permanently.
3. `FilterText` could be null on `default`, and it binds straight into a `TextField`.

## Verification

Unity MCP EditMode: **528 passed / 0 failed**, three times -- twice back-to-back on a persistent domain and again after rebasing onto master -- each against an assembly proven fresh by resolving a symbol added in the same edit. Baseline before the change was 470/0. `npm test` 400/400, `npm run validate:all` clean.

A full run including `...Tests.Editor.Allocations` shows 3 failures in `BenchmarkHarnessRobustnessTests` (`[Assert] Access version should be odd when acquiring lock`). These were **not** assumed pre-existing: checking out `master`, proving `DxMessagingMessageMonitorLiveView` absent from the loaded editor assembly, and re-running the same fixture gives the identical **112 passed / 3 failed**. It is a local host-editor artifact, unreachable from this change.

## Also here

- `scripts/llm.meta` was untracked. `scripts/llm/` shipped in #293 with a meta for `harness.js` but none for the directory, so a fresh Unity import regenerated it as an untracked file on every clone.
- Two UI Toolkit test-interaction facts added to `unity-editor-conventions`: a detached visual tree raises no `ChangeEvent`/`ClickEvent`, and a synthesized `ClickEvent` does not invoke `Button.clicked` (which is why every package button is wired with `RegisterCallback<ClickEvent>`).

## Follow-ups (not in this PR)

- A poll that changes the log rebuilds the body, so a scrolled-down reader loses scroll position. Pausing the recorder avoids it; incremental `ListView.RefreshItems` updating is the real fix.
- `.dx-prio`, `.dx-tip`, `.dx-danger` and `.dx-admonition__title` are the only classes left with no C# reference. Recorded as PLAN WS-2.4 rather than force-fit into a home they do not have.

---

> [!NOTE]
> **Medium Risk**
> Editor-only diagnostics UI with substantial new polling, bus-reset, and play-mode transition logic tied to emission sequence numbers; snapshot mode is additive but the window lifecycle changes are easy to get wrong under domain reload and fast emission rates.
> 
> **Overview**
> Adds **Live** mode to the Message Monitor: a **Live** button on the snapshot view swaps the window to a columnar log that **polls** the bus emission buffer (no runtime hooks), coalesces identical adjacent emissions, and reports **missed** counts when the ring overwrites before drain.
> 
> The live surface includes record/pause, Untargeted/Targeted/Broadcast chips, the same typed filter query as snapshot mode, a detail pane with selection pinned by **dispatch trace id** (not list index), and footer stats (shown, buffered, recorded, missed). `MessageMonitorLiveRecorder` handles drain bookkeeping; `DxMessagingMessageMonitorLiveView` renders the design-system recorder UI; the window runs a 250 ms pump and clears or rebases the log on play-mode transitions when the bus dispatch counter resets.
> 
> `DxMessagingEditorTheme` gains route-kind **dot** and **chip** helpers plus columnar recorder class names; `MessageMonitorEntry` now carries `TraceId` from emissions. Changelog and editor-conventions docs note detached UITK trees and `ClickEvent` vs `Button.clicked` for tests. Adds `scripts/llm.meta` for Unity import hygiene.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac671fc20aba64d7818b65f30345df242d5bac49. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


wallstop:
@cursor review

@copilot review

wallstop:
Addressed all three Cursor Bugbot findings in 6cd413bd, plus a related defect the first finding exposed.

**Play-mode clear refilled by drain (High)** — the root cause was broader than the play-mode path: `Clear()` rewound the cursor to 0, so *any* clear (including the Clear button) re-ingested everything the bus still had buffered on the very next poll. `Clear()` now keeps the cursor and drops only the rows, stats and loss baseline. A bus reset additionally drops the retained rows, since their `#N` dispatch labels would otherwise collide with the new run's.

**Live selection drifts on updates (Medium)** — selection is now keyed by the row's dispatch id instead of its list position, so it is stable across appends. A pinned row that is filtered away or ages out falls back to the newest row without losing the pin.

**Paused drain still snapshots bus (Medium)** — the drain now returns early when recording is off, instead of relying on a cursor that a paused recorder never advances.

New regression tests: `ClearingDoesNotRefillFromWhatTheBusStillHasBuffered`, `ABusResetRebasesTheCursorAndDropsThePreviousRun`, `APinnedRowKeepsItsDetailPaneAsNewerRowsArrive`, `APinnedRowThatIsFilteredAwayFallsBackWithoutLosingThePin`.

Local EditMode via the Unity MCP loop: 531 passed / 0 failed.

@cursor review

wallstop:
Fixed in f57acbed. This one was a genuine hole, and my own comment had gone stale on it.

`Ingest` can only infer a restarted dispatch sequence while the new run is still numbered *below* the cursor. Keeping the cursor across `Clear()` fixed the refill, but it left exactly the case you describe: a play-mode transition restarts the bus, the new run emits past the old cursor before the next 250 ms poll, and its opening emissions look already-drained. With the loss baseline reset they were not reported as missed either. The comment claimed the play-mode clear closed that gap; once `Clear()` stopped rewinding the cursor, it no longer did.

The sequence alone cannot close it, so the caller now announces it. `RebaseTo(busCursor)` empties the log and moves the cursor onto whatever counter the bus actually holds, and the window calls it on the **Entered\*** play-mode transitions rather than `Exiting*` — after the transition completes, so the counter it reads is the one that will produce the next emissions. That is correct either way:

- Bus reset: reports 0, so the whole new run is taken.
- Bus kept running (domain reload disabled): reports its live counter, so the backlog is correctly skipped and nothing refills.

`Clear()` is now just `RebaseTo(Cursor)`, and the inference in `Ingest` stays as a best-effort net for an unannounced `MessageBus.Reset()`, with its comment corrected to say so.

New tests: `RebasingOntoAResetBusTakesTheWholeNewRun` (the exact scenario you described), `RebasingOntoALiveBusSkipsEverythingBeforeIt`, `ANegativeRebaseTargetIsTreatedAsTheStartOfARun`.

Local EditMode via the Unity MCP loop: 534 passed / 0 failed.

@cursor review

wallstop:
Both fixed in 76c88d30.

**Inferred vs announced reset disagreed** — the inferred path in `Ingest` cleared rows and rewound the cursor but left `ObservedCount`/`MissedCount` alone, while `RebaseTo` reset them. Both now go through one `ResetLog`, so they cannot diverge; the footer no longer totals a run that is over. Covered by `AnInferredResetClearsTheSameStateAnAnnouncedOneDoes`.

**`DrainLiveRecorder` comment promised more than it delivered** — you're right that `Ingest` returns on an empty list before reaching reset detection, and `MessageBus.Reset` empties the buffer, so the log kept showing the previous run until the new one happened to emit. The drain now compares the bus counter against the cursor itself and calls `RebaseTo` immediately, which is what the comment always claimed. The idle shortcut is now an explicit `==` branch rather than something the reset case fell through.

Local EditMode via the Unity MCP loop: 535 passed / 0 failed.

@cursor review

wallstop:
Fixed in ac671fc2 — this was the sharpest of the findings.

You're right that `EmissionId` names the most recent emission **already in the ring**, so adopting it stepped over everything the new run had produced. `EnteredPlayMode` fires after `Awake`/`OnEnable`/`Start` have emitted, so the live log permanently lost exactly the startup traffic it exists to show. The drain's restart branch had the same hole once the buffer had refilled.

Rather than patch the arithmetic, I removed the ability to express the wrong thing. The cursor now only ever moves in the two directions that are actually meaningful, and neither takes a counter:

- `ResetForNewBusRun()` rewinds to the start of a run, so everything the bus currently holds is drained. Used when the counter restarted.
- `Clear()` keeps the cursor — which, on a bus that did *not* restart, already sits exactly on the boundary between the old session and the new one.

The play-mode handler picks between them on whether the bus counter has fallen below the cursor. Both branches now keep the new session's opening emissions: a reset bus drains its whole buffer, and a bus that kept running (domain reload disabled) drains everything past the boundary.

New test `ANewSessionOnARunningBusKeepsTheBoundaryAndTakesWhatFollowsIt` covers the domain-reload-disabled path explicitly, asserting the Awake/Start emissions survive; `ResettingForANewBusRunTakesEverythingTheNewRunHasAlreadyBuffered` covers the reset path.

Local EditMode via the Unity MCP loop: 534 passed / 0 failed.

@cursor review

Issue #304, opened 2026-07-29
[Feature]: Decide adopt-or-prune for the last four unused .dx-* classes


wallstop:
## Context

`Editor/Theme/DxMessagingTheme.uss` was migrated from the design-system spec wholesale, so it described a larger surface than the windows actually rendered. That gap has been closed in stages:

- The subscriptions section (#295) adopted `.dx-inspector` / `.dx-sub*`.
- The live Message Monitor (#302) adopted `.dx-record`, `.dx-chip`/`.dx-filter`, `.dx-list-header` + `.dx-col-*`, `.dx-row`/`.dx-row--alt`/`.dx-row__*`, `.dx-detail`/`.dx-detail__*`, `.dx-kv*` and `.dx-footer*`.

## What is left

Exactly four classes still have no C# reference:

| Class | Notes |
| --- | --- |
| `.dx-prio` | A small accent-coloured priority badge. The subscriptions section already renders a priority, but as plain text inside its meta line (`{registrationType} \| priority {n} \| {calls}`). Promoting that number to a badge is the obvious adoption. |
| `.dx-tip` | Admonition variant. No surface currently has a "tip" state. |
| `.dx-danger` | Admonition variant. Closest candidate is the base-call violation, but that is deliberately styled as `.dx-warning` today, and changing its severity is a product decision, not a styling one. |
| `.dx-admonition__title` | Only reachable as a descendant of an admonition variant; blocked on the two above. |

## Ask

Decide per class: adopt into a real surface with a contract test, or prune from the stylesheet. **Do not leave them dead** -- an unused class is drift that silently grows.

Verify with:

```
FILES=$(find Editor Tests -name '*.cs')
for c in dx-prio dx-tip dx-danger dx-admonition__title; do
  echo "$c : $(grep -l -- "$c" $FILES 2>/dev/null | wc -l) files"
done
```

Recorded in PLAN.md as WS-2.4.

wallstop:
Decided and shipped in #306 (`ffef8241`) -- and the question can no longer reopen quietly.

| Class | Decision |
| --- | --- |
| `.dx-prio` | **Adopt.** The subscriptions row's priority is now a badge instead of prose buried in the meta line. |
| `.dx-danger` | **Adopt.** Carries the live Monitor's "log has gaps" notice -- the one condition that makes every count on screen wrong. |
| `.dx-admonition__title` | **Adopt.** Titles that same notice. |
[...]

Issue #303, opened 2026-07-29
[Feature]: Live Message Monitor loses scroll position on each poll


wallstop:
## Context

The live Message Monitor shipped in #302. Its poll loop rebuilds the whole body whenever the log changed:

`DxMessagingMessageMonitorWindow.PumpLive` -> `RenderLiveBody` -> `DxMessagingMessageMonitorLiveView.RenderBody`, which does `body.Clear()` and rebuilds the `ListView`.

## Problem

A reader scrolled down into older rows loses their scroll position every time a new emission arrives (up to four times a second on a busy scene). Selection itself is stable -- #302 keys it by dispatch id -- but the viewport is not.

## Workaround that exists today

Turning the **Record** toggle off freezes the log: the drain stops, `Revision` stops moving, and the body stops rebuilding. That is a reasonable answer for "I want to study this one message", and it is what the toggle is for.

## The real fix

Update the list incrementally instead of rebuilding it:

- Keep the `ListView` instance across renders, reassign `itemsSource`, and call `RefreshItems()` rather than clearing the body.
- Rebuild the header/footer/detail separately from the list, since those are cheap and do not hold scroll state.
- Per UITK guidance, auto-scroll (if added) needs a **deferred** `ScrollToItem`; it no-ops when called directly after `RefreshItems`.

## Scope

Editor-only, `Editor/Windows/DxMessagingMessageMonitorLiveView.cs` plus the render path in `DxMessagingMessageMonitorWindow`. Recorded in PLAN.md as an M4 item.

wallstop:
Fixed in #306 (`ffef8241`). The live body is now four fixed slots and each is updated independently, so a poll no longer throws a scrolled reader back to the newest row.

- `EnsureBodyState` calls `body.Clear()` exactly once, on first render, and stashes the slots in `body.userData`; every later render reuses them.
- `UpdateList` refills the same `List<>` the `ListView` was built around and calls `RefreshItems()`. Reassigning `itemsSource` would be a collection reset, so it deliberately does not.
- Selection is only re-applied when it actually moved, which also keeps the poll allocation-free on an unchanged pin.
- The detail pane is likewise kept, keyed on the selected row's dispatch range and coalesced count so folding into a pinned row still refreshes it.

[...]

Issue #301, opened 2026-07-29
[Bug]: Improve Styling on GitHub Pages


wallstop:
### Description

The bar beneath the pages is a bit odd - it would be cool if it had links, or was rendered full-width.

<img width="1559" height="935" alt="Image" src="https://github.com/user-attachments/assets/75a90fb7-d14c-489a-966c-30798e2414bb" />

Everything else has full borders, this one just has top and bottom. It is also left-aligned instead of spread with spacing, leaving tons of space on the right.

### Steps to Reproduce

Visit https://ambiguous-interactive.github.io/DxMessaging/

### Expected Behavior

Great layout

### Actual Behavior

Decent layout

### Unity Version

2021.3 LTS

### Package Version

3.2.1

### Platform

- [ ] Editor
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] iOS
- [ ] Android
- [ ] WebGL
- [ ] Other

wallstop:
Fixed in #306 (`ffef8241`). The footer meta bar now carries links and spreads properly:

- `docs/overrides/partials/copyright.html` overrides the theme partial to render a `.dxm-footer-links` nav with Repository, Releases, Changelog, Report an issue, and License -- deliberately the destinations the site nav does not already carry.
- `docs/stylesheets/extra.css` sizes `.md-footer-meta .md-copyright` to `width: auto` so Material's `space-between` can do its job: copyright at one end, links and social at the other, wrapping on narrow screens. Previously the copyright claimed the full width, which is what pushed the single social icon onto a second left-aligned line and left the right-hand side empty.

Reopen if the bar still reads wrong to you at some viewport -- the layout change is the part worth re-checking visually.

wallstop:
We didn't fix this issue at all

<img width="1375" height="267" alt="Image" src="https://github.com/user-attachments/assets/f12d31b9-d00f-4185-b963-717f4eaad8e7" />

wallstop:
This should show:

10+ years in production
MIT License (linked)
zero dependencies
O(1) routing
zero allocations
sources (with links)

And be properly styled

Issue #299, opened 2026-07-29
[Bug]: GitHub Pages (Light Mode?) Is Super Slow For Some People


wallstop:
### Description

Clicking between different pages causes very slow load times for some people. It might be light-mode specifics.

### Steps to Reproduce

Open the latest GitHub pages. Click around.

### Expected Behavior

Super fast

### Actual Behavior

Slow.

### Unity Version

2021.3 LTS

### Package Version

3.2.1

### Platform

- [ ] Editor
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] iOS
- [ ] Android
- [ ] WebGL
- [x] Other

### Additional Context

Here's a video repro

https://github.com/user-attachments/assets/4d8eef68-c905-4bc3-bffe-5f938be160c3

wallstop:
We still appear to be serving extremely large payloads - can we significantly minimize them, across all links and pages? Do we really need a 3MB payload for mermaid?

wallstop:
Fixed in #306 (`ffef8241`). The per-page mermaid payload is gone: `mkdocs.yml` now feeds Material's own mermaid integration through the `superfences` custom fence, which loads mermaid only on the pages that actually contain a diagram and themes it from the `--md-mermaid-*` custom properties. There is no `extra_javascript` entry left, so the 3.5 MB bundle that used to ship to 48 of 49 pages is no longer requested at all.

That bundle was also what pushed tracked JS over its budget and turned master red, which is tracked separately as #307.

PR #306, opened 2026-07-29
fix: unblock master CI by deleting the per-page mermaid payload (#299), plus M4 Monitor and design-system polish

wallstop:
## Why master is red, and the fix

`validate:js-loc-budget` fails on master: tracked JS is 17535 lines against a 17500 ceiling.

Neither change that landed did anything wrong. #294 added +42 lines (a probe regression fix plus its
test) on top of a base at 17457; master had meanwhile moved to 17493. Each was individually under the
ceiling, PR CI had evaluated the stale base, and the merge put them over. The budget's own header says
to delete rather than raise, so this deletes -- and the thing worth deleting turned out to be the
cause of an open bug.

## #299: GitHub Pages is slow

`extra_javascript` pulled `https://unpkg.com/mermaid@11/dist/mermaid.min.js` -- **3,565,102 bytes raw,
948 KB gzipped** -- into **48 of the site's 49 pages**. **Six** of them render a diagram.

`docs/javascripts/mermaid-config.js` (261 lines) then initialized mermaid a *second* time with a
conflicting config and re-rendered the same `pre.mermaid` elements Material had already claimed, on
every `document$` emission and every palette toggle.

The `pymdownx.superfences` custom fence that feeds Material's own diagram support was already
configured. Material loads mermaid lazily -- only on pages that contain a diagram -- and themes it
from 26 `--md-mermaid-*` custom properties that default to `--md-accent-fg-color`,
`--md-code-fg-color` and `--md-default-bg-color`, all of which `extra.css` already redefines per
scheme. So deleting the entry and the script is the whole fix, and diagrams keep the DxMessaging
palette in both light and dark with no diagram-specific CSS.

Verified against a local `mkdocs build --strict`:

| | before | after |
|---|---|---|
| pages requesting mermaid up front | 48 / 49 | 0 |
| pages carrying `<pre class="mermaid">` | 6 | 6 |
| tracked JS | 17535 | 17271 |

Space Grotesk also moved out of an `@import` inside `extra.css` into a sibling `extra_css` link: an
`@import` cannot start downloading until the importing sheet is fetched and parsed, which serialized
two render-blocking requests on first load. `docs/javascripts/csharp-highlight.js` was an empty IIFE
no template referenced; it went with the directory.

Diagram colours change: they now follow the site accent instead of the deleted script's five-hue
palette. That is intentional -- one palette, and it is the one the rest of the page uses -- but it is
a visual change worth a look.

## #303: the live Monitor lost your place every poll

`RenderBody` cleared the body and rebuilt it, so every poll that changed the log built a new
`ListView`, and a new list starts at the top. A reader who had scrolled into older rows was thrown
back to the newest row up to four times a second.

The body is now four fixed slots and a render updates each independently. The list is kept and
refreshed: its item source is the same `List<>` the view was built around, refilled in place so the
update stays a data refresh rather than a collection reset, and the row binding reads the current
rows, view state and callbacks off shared state instead of capturing them -- so a refreshed list never
renders a stale row set or reports a click to a host that has been replaced. The detail pane is kept
under the same reasoning, because it carries a scrollable stack trace.

Two things deliberately not done: `ListView.scrollOffset` is untouched (unproven on the 2021.3 floor,
and nothing else in this package uses it), and the residual newest-first drift -- arriving rows still
push older ones down under a fixed pixel offset -- is inherent to the ordering rather than to the
rebuild.

## #304: the last dead theme classes, and the invariant

#304 listed four. Hand-auditing is exactly why its list was incomplete: `.dx-sep` was dead too.

Adopted `.dx-prio` (the subscriptions row promotes priority out of its meta sentence into the accent
badge), `.dx-danger` + `.dx-admonition__title` (the live Monitor now says when its log has gaps
instead of leaving a non-zero `missed` stat to be noticed -- a hole makes every count on screen wrong,
which is louder than the warnings elsewhere in the package), and `.dx-sep` (the live toolbar ran four
unrelated control groups together). Pruned `.dx-tip`: no surface has a tip severity, and the surfaces
that would want one are empty states, which have `.dx-empty`.

The real fix is `EveryStylesheetClassIsRenderedByAnEditorSurface`, which parses both stylesheets with
comments stripped and fails on any `.dx-*` class with no quoted C# reference. 66 declared, none dead.
A class added to the sheet but never rendered now fails, and so does a rendered class whose last
reference was deleted.

## #301: the footer meta bar

The theme sets `.md-copyright { width: 100% }` so its "Made with Material for MkDocs" line can sit
under the copyright. This site sets `generator: false`, so that width only pushed `.md-social` onto a
second line, left-aligned, with the rest of the row empty. Sizing the copyright to its content lets
`space-between` work, and `partials/copyright.html` adds the destinations the nav does not carry
(repository, releases, changelog, report an issue, license -- all five verified 200). Full-bleed
content was deliberately not adopted: the bar shares `.md-grid` with the header, tabs and body, so
widening only it would misalign the one element the report is about.

## Also

The SourceGenerators test csproj was committed with CRLF in the index while `.gitattributes` marks
`*.csproj` as `text eol=crlf`, so every fresh checkout showed it modified. Renormalized.

## Verification

- MCP EditMode against the host editor (6000.4.6f1): **545 passed / 0 failed**
  (`.artifacts/unity-mcp/session-173-editmode-2.json`); the live-view fixture alone is 40 / 0.
- `npm test`: 401 passed / 0 failed.
- `npm run validate:all`: every gate green, `validate-js-loc-budget: OK (17271/17500)`.
- `mkdocs build --strict`: clean, and the before/after table above was measured from its output.

## Follow-up filed

Two PRs each individually under the JS ceiling can still merge over it, because PR CI evaluates the
base as it was when the run started. Filed separately rather than papered over here.

## Review round

Cursor Bugbot found one real issue and it is fixed: the gap notice's body label carried
`.dx-empty__body`, which caps width at 260px and centers text -- right for a centered empty state,
wrong for a full-width footer notice. The label now carries no class, matching this package's other
admonitions where `.dx-admonition` owns the padding, and the test asserts the empty-state class is
absent so the convenient-but-wrong choice cannot come back. Bugbot's next pass came back clean.

Copilot could not review: the requesting account has reached its review quota. Flagged rather than
worked around.

Two more findings came from re-reading the diff rather than from a bot:

- The two colored `GUIStyle`s in the component inspector were cached with `??=`. Harmless while their
  colors were skin-independent constants, not harmless once they are not: switching the editor theme
  repaints without reloading the domain, so a cached style kept painting the other skin's color. They
  are now rebuilt when the skin they were built for stops matching.
- `theme.custom_dir` points inside `docs_dir`, so MkDocs was copying every file under
  `docs/overrides/` into the published site -- raw Jinja templates, the `.csharp` snippets, and their
  Unity `.meta` sidecars. Pre-existing, but adding the footer partial put one more file there and made
  it visible. `overrides/` joins `ops/` and `runbooks/` in `exclude_docs`, and the build confirms the
  templates still work: the home page renders its custom hero, snippets still inline, and the footer,
  diagram and mermaid invariants above all still hold.

## Final state

**42 checks: 33 success, 8 skipped, 1 failure.** The failure is
`copilot-pull-request-reviewer`, which reports that the requesting account has reached its review
quota; it is not a code result and is not one of the required gates.

Both aggregate gates are green -- `CI Success` and `Unity CI Success` -- and all nine Unity legs pass:

| | editmode | playmode | standalone |
| --- | --- | --- | --- |
| 2021.3.45f1 | pass | pass | pass |
| 2022.3.45f1 | pass | pass | pass |
| 6000.3.16f1 | pass | pass | pass |

`Cursor Bugbot` ends at success with no outstanding findings. `mergeable_state` is `clean`.


wallstop:
bugbot run

wallstop:
Bugbot's finding is fixed in 727a1f56: the gap notice's body label carried `.dx-empty__body`, which caps width at 260px and centers text -- right for a centered empty state, wrong for a full-width footer notice. The label is now unclassed, matching this package's other admonitions where `.dx-admonition` owns the padding, and `ALossyLogSaysSoInsteadOfLeavingItToTheMissedStat` asserts the empty-state class is absent and the label wraps, so the convenient-but-wrong class cannot come back.

Copilot reported it could not review because the requesting account has hit its review quota; flagging rather than working around it.

bugbot run

Issue #307, opened 2026-07-29
ci: a stale PR base can merge two changes that are each under the JS budget but together over it


wallstop:
## Problem

Master's `CI Success` went red on `validate:js-loc-budget` with no bad change involved.

| commit | tracked JS |
| --- | --- |
| `7477cb6a` (base #294 was cut from) | 17457 |
| `f09dfc08` (master when #294 merged) | 17493 |
| `146dd9f3` (#294 merged, +42) | **17535** (ceiling is 17500) |

#294 added 42 lines. Against the base it was tested on that is 17499 -- one line under the
ceiling, and its checks passed honestly. Master had moved to 17493 in the meantime, and GitHub
does not re-run a pull request's checks when the base branch moves unless the branch is
required to be up to date before merging. So the passing run had evaluated a base that no
longer existed at merge time, and two individually-legal changes produced an illegal total.

This is not specific to the JS budget. Any whole-repository invariant -- a total, a count, a
uniqueness constraint -- can be broken by two PRs that are each fine alone. The budget just
happens to be the one that reports it.

## Why it was not obvious

The failure surfaced only on the push-to-master run, after the merge. Nothing on either pull
request indicated a problem, and neither author did anything wrong. The fix in #306 was to
delete JS (which the budget's own header asks for), but that only clears the current overshoot;
the next concurrent pair can do it again.

## Options

1. **Require branches to be up to date before merging** on the ruleset that gates `master`.
   Closes the hole exactly: the branch has to be rebased, which re-runs the checks against the
   real base. Cost is a rebase-and-wait per PR when master is busy, which for this repository's
   merge rate is small.
1. **Keep deliberate headroom in the budget.** #306 leaves 229 lines. This does not close the
   hole, it just makes the collision unlikely -- and it silently weakens the constraint the
   budget exists to enforce.
1. **Merge queue.** Correct and automatic, but it serializes merges behind the Unity matrix,
   which is the expensive leg on this repository.

Option 1 is the recommendation. Option 2 is what is in place by accident today and should not
be relied on.

## Acceptance criteria

- A pull request whose base has moved cannot merge on a stale passing run for the checks that
  enforce repository-wide invariants.
- `validate:js-loc-budget` failing on `master` again for this reason is treated as a gate bug,
  not as a budget problem.

wallstop:
Confirmed against the live ruleset while working #309. `Required CI - Unity Tests (default branch)` (id `17663217`) has:

```json
[...]

Issue #289, opened 2026-07-21
[Feature]: Investigate Perf Wins From Custom HashTables


wallstop:
### Feature Description

See if we can learn anything from the Java world: https://sooniln.github.io/posts/hashmap-benchmarks-2026/ and replace any of our Dictionary uses with something faster?

### Use Case / Motivation

More perf

### Proposed Solution

Red/green+perf test implementations

### Alternatives Considered

Don't do it

wallstop:
## Measured, on the real dispatch path

Investigated through the Unity MCP loop against the host editor (Unity 6000.4.6f1, Mono in-editor, warmed then one continuous window per variant, 4M lookups per cell).

### First, where a hash lookup actually happens per emit

Most of the bus is already dictionary-free. `MessageCache<TValue>` is a **sparse array indexed by a compact per-message-type integer** (`MessageHelperIndexer<TMessage>`), so every type-keyed lookup is array indexing, not hashing. Steady-state dispatch then walks a frozen `FlatDispatchEntry<TMessage>[]` with plain `entries[i]`.

That leaves exactly **one** hash lookup on the per-emit path, and only for two of the three kinds:

```csharp
// MessageBus.TargetedBroadcast, fast lane
Dictionary<InstanceId, HandlerCache<int, HandlerCache>> fastTargeted = plan.contextHandle;
if (fastTargeted != null && fastTargeted.TryGetValue(target, out HandlerCache<int, HandlerCache> fastSorted) ...
```

`SourcedBroadcast` has the same shape. **`UntargetedBroadcast` has no hash lookup at all** -- it reads `plan.scalarHandle` directly. So this is a targeted/broadcast question only, and the Java HashMap benchmarks in the linked article are about a structure this codebase has already mostly designed away.

### The measurement

`InstanceId` is a `readonly struct` holding `int _id` **plus a managed `UnityEngine.Object Object`**, with `Equals`/`GetHashCode` reading only `_id`. Three key shapes, identical identity semantics, 64 targets, strided realistic ids:

| key shape | ns / lookup | vs today |
| --- | --- | --- |
| `InstanceId`-shaped struct (int + `Object` ref) -- **today** | ~23.3 | -- |
| struct carrying **only** the int | ~25.2 | none (slightly worse) |
| raw `int` | ~12.3 | **1.9x faster** |
| open-addressed `int` table (linear probe, pow2 mask) | ~4.9 | **4.8x faster** |

[...]

PR #309, opened 2026-07-30
ci: gate committed line endings and stop the watchdog cloning master

wallstop:
## Why

The **Stuck Job Watchdog** failed 13 runs straight on 2026-07-29 (18:10Z through 23:54Z), every time aborting at the same place:

```text
error: Your local changes to the following files would be overwritten by checkout:
    SourceGenerators/.../WallstopStudios.DxMessaging.SourceGenerators.Tests.csproj
Please commit your changes or stash them before you switch branches.
```

## Root cause

That blob was committed **holding CRLF bytes** (2010 B, 45 CR) while `.gitattributes` declares `*.csproj text eol=crlf`:

| commit | blob size | CR |
| --- | --- | --- |
| `a9e1ff30` (Dependabot, 18:09Z) | 2010 | 45 |
| `378a7d7f` | 2010 | 45 |
| `ffef8241` (23:58Z) | 1965 | 0 |

Under an explicit `text` attribute git always normalizes on the clean side. Git also skips the LF-to-CRLF smudge when the blob already holds CRLF, so the checked-out file is byte-identical to the blob -- and the clean filter still converts it back to LF and hashes to something else. A pristine clone therefore reports the file modified with no way to make it un-modified. The watchdog cloned the default branch and then switched to the `watchdog-state` orphan tree, which must delete every default-branch file, and git refuses to delete a modified one.

The window opened with `a9e1ff30` and closed with `ffef8241`, which happened to renormalize the file as a side effect. Nothing prevents the next Dependabot bump from reopening it.

Reproduced deterministically in a scratch repo (same `.gitattributes` ordering, blob written with `hash-object --no-filters`), which emits the identical error, and confirmed the fix makes it unreachable.

## What changed

**1. `ci.yml` gains a `line-endings` job.** It runs `git add --renormalize .` and fails, annotating each offending path.

Nothing else could have caught this. An ordinary diff is empty in a pristine clone, so only `--renormalize` sees it. The `csharpier` job already ran `git add --renormalize .` -- but follows it with `git reset --hard`, so it repairs the worktree for the formatter and discards the evidence. And Dependabot commits through the API, so it never runs the pre-commit `mixed-line-ending` hook; the gate has to live in CI.

The job is deliberately unfiltered (every tracked text file is in scope, so there is no useful path filter and therefore no `changes` output to fail closed against) and is wired into `CI Success`, so branch protection is unchanged.

**2. The watchdog no longer materializes the default branch.** It clones the state branch directly (`--single-branch --branch "${STATE_BRANCH}"`) instead of cloning the default branch and switching away from it. The watchdog only ever reads and writes `.watchdog-state/`, so no default-branch file is written and none can block it -- the failure class is unreachable rather than merely fixed. The bootstrap path now builds the orphan with `git init` instead of emptying a default-branch checkout, which is also strictly less work.

## Tests

Two guards in `scripts/__tests__/ci-aggregate-workflow.test.js`:

- the gate must **fail** on drift rather than repair it (`exit 1` in `line-endings`, `reset --hard` in `csharpier`, never the reverse);
- the watchdog's single `clone` must stay state-branch-scoped, with no cross-tree `checkout -B`.

Verified locally: `node --test scripts/__tests__/ci-aggregate-workflow.test.js scripts/__tests__/auto-commit-fetch-force-refspec.test.js` -> 19/19; `actionlint` and `yamllint` clean on both workflows; `git add --renormalize .` on a fresh clone of `master` reports no drift, so the new gate is green on the current tree.

---

> [!NOTE]
> **Medium Risk**
> Touches CI gating and git attribute semantics repo-wide; behavior is intentional and test-guarded, but mis-merging `.gitattributes` back to bare `text` could reintroduce silent dirty clones.
> 
> **Overview**
> Fixes **Stuck Job Watchdog** failures where a pristine clone looked dirty after Dependabot/API commits left CRLF blobs that disagreed with `.gitattributes`, blocking branch switches.
> 
> **`.gitattributes`** switches every text rule from bare `text` to **`text=auto`** (same `eol` targets) so already-CRLF blobs are not forced through clean-side normalization that makes clones appear modified with no fixable edit.
> 
> **CI** adds an unfiltered **`line-endings`** job: `git add --renormalize .` then **`exit 1`** if the index differs, with per-file annotations. It is wired into **`CI Success`**. The **csharpier** job still renormalizes and **`reset --hard`** for formatting only—documented as repair, not the drift gate.
> 
> **Stuck Job Watchdog** clones **`watchdog-state`** directly (`--single-branch --branch`) and bootstraps with **`git init`** instead of checking out the default branch and switching trees, so default-branch line-ending drift cannot block state updates.
> 
> **Tests** in `ci-aggregate-workflow.test.js` lock in: gate vs repair behavior, no bare `text` in `.gitattributes`, and state-branch-only cloning. **`.llm/context.md`** documents the required Unity MCP **`resultPath`** package prefix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17602286a50647cb42cd03b3f828b4746626bf26. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,42 +3,65 @@
 #
 # Default to LF line endings for all text files for better cross-platform
 # compatibility and CI/CD consistency. C# files use CRLF per .NET conventions.
+#
+# Every rule below says `text=auto`, never a bare `text`. Do NOT "tighten" them
+# back: `text` is what turns a cosmetic problem into an outage.
+#
+# A blob can reach this repository already holding CRLF, because commits made
+# through the GitHub API (Dependabot, Copilot) never run a git client and so
+# never apply these attributes. Under a bare `text`, git normalizes such a blob
+# on the clean side unconditionally while skipping the smudge (the blob already
+# has CRLF) -- so the checked-out file matches the blob byte for byte, the clean
+# filter hashes to the LF form, and a PRISTINE CLONE REPORTS THE FILE MODIFIED
+# with no edit that can make it clean. Any consumer that switches branches or
+# gates on a clean tree then breaks: that is what failed the Stuck Job Watchdog
+# 13 runs straight on 2026-07-29, and the same blob had been silently
+# oscillating LF/CRLF through Dependabot bumps since 2025-11.
+#
+# `text=auto` leaves an already-CRLF blob alone instead, so the clone stays
+# clean and the switch works, while still normalizing everything a git client
+# commits and still delivering the eol below to the working tree (verified
+# identical for all tracked files). It makes the condition harmless; it cannot
+# make it impossible, because attributes are instructions to a git client. The
+# `line-endings` job in ci.yml is what still reports the drift
+# (`git add --renormalize .` detects it under either form), so the index does
+# not quietly accumulate CRLF.
 ###############################################################################
 * text=auto eol=lf
 
 # C# and .NET project files use CRLF
-*.cs        text eol=crlf
-*.csproj    text eol=crlf
-*.sln       text eol=crlf
-*.props     text eol=crlf
+*.cs        text=auto eol=crlf
+*.csproj    text=auto eol=crlf
+*.sln       text=auto eol=crlf
+*.props     text=auto eol=crlf
 
 # Explicit LF entries for common text file types.
 # Other text files also use LF via the default rule: * text=auto eol=lf
-*.json      text eol=lf
-*.yaml      text eol=lf
-*.yml       text eol=lf
-*.md        text eol=lf
-*.xml       text eol=lf
-*.uxml      text eol=lf
-*.uss       text eol=lf
-*.shader    text eol=lf
-*.hlsl      text eol=lf
-*.compute   text eol=lf
-*.cginc     text eol=lf
-*.asmdef    text eol=lf
-*.asmref    text eol=lf
-*.meta      text eol=lf
-*.ps1       text eol=lf
-*.txt       text eol=lf
-*.js        text eol=lf
-*.mjs       text eol=lf
+*.json      text=auto eol=lf
+*.yaml      text=auto eol=lf
+*.yml       text=auto eol=lf
+*.md        text=auto eol=lf
+*.xml       text=auto eol=lf
+*.uxml      text=auto eol=lf
+*.uss       text=auto eol=lf
+*.shader    text=auto eol=lf
+*.hlsl      text=auto eol=lf
+*.compute   text=auto eol=lf
+*.cginc     text=auto eol=lf
+*.asmdef    text=auto eol=lf
+*.asmref    text=auto eol=lf
+*.meta      text=auto eol=lf
+*.ps1       text=auto eol=lf
+*.txt       text=auto eol=lf
+*.js        text=auto eol=lf
+*.mjs       text=auto eol=lf
 
 # Shell scripts must use LF for proper execution on Unix systems
-*.sh        text eol=lf
-*.bash      text eol=lf
-*.zsh       text eol=lf
-*.ksh       text eol=lf
-*.fish      text eol=lf
+*.sh        text=auto eol=lf
+*.bash      text=auto eol=lf
+*.zsh       text=auto eol=lf
+*.ksh       text=auto eol=lf
+*.fish      text=auto eol=lf
 
 
 ###############################################################################
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -296,6 +296,9 @@ jobs:
         with:
           persist-credentials: false
 
+      # Repairs the worktree so CSharpier sees one canonical form; it is NOT the
+      # drift gate. The `line-endings` job is what fails on committed bytes that
+      # disagree with .gitattributes.
       - name: Normalize line endings
         if: ${{ needs.changes.outputs.csharpier != 'false' }}
         run: |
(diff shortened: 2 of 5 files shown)


wallstop:
bugbot run

wallstop:
## One thing found while writing this, which changes what the PR is worth

The `line-endings` job is wired into `CI Success` -- and **`CI Success` was not a required status check.**

Ruleset `Required CI - Unity Tests (default branch)` (`17663217`) required `Unity CI Success` plus 14 *individual* static job names:

```text
Lint repository Markdown      Check C# formatting        dotnet tests
Check JSON/.asmdef formatting Check spelling             Validate banner SVG
Check llms.txt is up-to-date  Prettier and yamllint      Lint GitHub Actions workflows
Script tests (ubuntu-latest)  Script tests (macos-latest) Script tests (windows-latest)
Validate Documentation Build  Lint docs links
```

`.llm/context.md` says the opposite in as many words: *"Do not require individual static job names ... after the aggregate ruleset switch."*

So the gate in this PR would have shipped **inert** -- red at merge time with nothing to block the merge. `CI Success` has been added to the ruleset (additive; nothing removed), which this PR's own run confirms is safe: `CI Success` reports on `pull_request` as well as `push`, and it cannot be tripped by a skip -- every job it needs is `if: ${{ always() }}` at job level and `alls-green` runs with `allowed-skips: ""`.

Removing the 14 now-redundant names is #310, deliberately separate: adding a gate is strengthening and safe in passing, removing 14 required names reads as weakening even where the aggregate makes it equivalent.

## Verification on this PR's run

- **`Check committed line endings` -> success**, and `CI Success` -> success with it.
- `Script tests` green on all three OSes (403/0 locally, including the two new contract tests).
- `actionlint` + `Prettier and yamllint` green on both changed workflows.

## Related

- #310 (filed) -- remove the redundant individual required checks.
- #307 -- commented with the confirmed `strict_required_status_checks_policy: false`. Not addressed here: it is a merge-workflow cost decision, not a code fix.
- #289 -- unrelated to this PR, but measured this session and written up with data; the premise turned out to be wrong.

Copilot's review returned *"unable to review ... the user who requested the review has reached their quota limit"*, so that check's failure is a quota result rather than a finding.

wallstop:
bugbot run

wallstop:
## This was not a one-off. It has been oscillating for nine months.

I traced the blob's full history rather than stopping at the commit that broke the watchdog. Every commit that has ever touched `SourceGenerators/...Tests.csproj`, with its committed CR count and author:

| date | CR | author | commit |
| --- | --- | --- | --- |
| 2025-10-31 | 0 | Eli Pinkerton | Full Unity Safeguards, Minor Bugfixes (#80) |
| 2025-11-25 | **20** | dependabot[bot] | Bump Microsoft.NET.Test.Sdk 18.0.0 -> 18.0.1 (#101) |
| 2026-01-17 | **20** | dependabot[bot] | Bump NUnit3TestAdapter 5.2.0 -> 6.0.0 (#117) |
| 2026-01-18 | **20** | dependabot[bot] | Bump NUnit3TestAdapter 6.0.0 -> 6.1.0 (#121) |
| 2026-01-21 | 0 | Eli Pinkerton | **Update EOL settings (#125)** |
| 2026-02-15 | **20** | dependabot[bot] | Bump coverlet.collector 6.0.4 -> 8.0.0 (#168) |
| 2026-02-16 | 0 | Copilot | Fix changelog display in UPM |
| 2026-02-18 | **20** | dependabot[bot] | Bump NUnit 4.4.0 -> 4.5.0 (#176) |
| 2026-02-24 | **20** | dependabot[bot] | Bump Microsoft.NET.Test.Sdk 18.0.1 -> 18.3.0 (#177) |
| 2026-02-26 | 0 | dependabot[bot] | Bump actions/upload-artifact 6 -> 7 (#178) |
| 2026-03-08 | **20** | dependabot[bot] | Bump NUnit 4.5.0 -> 4.5.1 (#180) |
| 2026-03-15 | 0 | dependabot[bot] | Bump release-drafter 6 -> 7 (#182) |
| 2026-03-17 | **20** | dependabot[bot] | Bump coverlet.collector 8.0.0 -> 8.0.1 (#185) |
| 2026-03-22 | **20** | dependabot[bot] | Bump NUnit3TestAdapter 6.1.0 -> 6.2.0 (#187) |
| ... | | | |
| 2026-07-29 | **45** | dependabot[bot] | Bump Microsoft.NET.Test.Sdk 18.7.0 -> 18.8.1 (#286) |
| 2026-07-29 | 0 | (renormalized) | `ffef8241` |

Every CRLF reintroduction is a `dependabot[bot]` commit that edits the file. Every clearing is a human commit or a Dependabot PR whose branch happened to get an autofix commit -- the `prettier-autofix` / `csharpier-autofix` workflows run `git add --renormalize` and commit, so they have been silently repairing this by accident for nine months.

**Note the 2026-01-21 row: "Update EOL settings (#125)".** This has already been fixed once by hand. It did not hold, because nothing gated it.

## Scope, checked rather than assumed

The other two project files have **never** carried CRLF:

| file | CR, every commit |
| --- | --- |
| `...SourceGenerators.csproj` | 0 (8/8 commits back to 2025-06) |
| `...Analyzer.csproj` | 0 (4/4 commits) |

Because Dependabot never edits them -- only the Tests csproj carries `PackageReference` versions. **The one file Dependabot edits is the one file that oscillates**, which confirms the mechanism: API commits bypass the pre-commit `mixed-line-ending` hook, and nothing downstream ever failed on the result.

## Why this changes the justification

Each CRLF period was a live window in which a pristine clone of `master` was dirty and any cross-tree branch switch aborted. That is at least **eight** such windows since October 2025. The watchdog outage on 2026-07-29 was not the defect appearing; it was the first time a workflow existed that did such a switch on a five-minute cron, so the pre-existing condition finally became visible.

The repository's own rule is not to add checks unless they prevent a real, recurring problem. This is eight recurrences in nine months, one previous hand-fix that regressed, and an accidental self-healing mechanism that made it invisible in between. It also means the gate had to be repo-wide (`git add --renormalize .`) rather than scoped to the known file: the moment a new test project gains `PackageReference` bumps, it inherits the same behaviour.

wallstop:
bugbot run

wallstop:
## Can this be enforced via `.gitattributes`? Mostly yes -- added in `17602286`

Measured rather than reasoned about, with a CRLF blob committed via `hash-object --no-filters` to reproduce the Dependabot-via-API shape:

| behaviour | `text eol=crlf` (before) | `text=auto eol=crlf` (now) |
| --- | --- | --- |
| pristine clone status | ` M bad.csproj` -- spuriously dirty | **clean** |
| cross-tree branch switch | **aborts** | **works** |
| `--renormalize` still sees drift | yes | **yes** |
| LF blob -> worktree eol | `i/lf w/crlf` | `i/lf w/crlf` |
| `git add` of a CRLF worktree file -> blob | LF | LF |

A bare `text` normalizes an already-CRLF blob on the clean side unconditionally while skipping the smudge, so the file matches the blob byte for byte and the clean filter hashes to the LF form -- dirty with no edit that can fix it. **`text=auto` leaves such a blob alone**, so the clone stays clean and the switch works.

### The division of labour, precisely

The two layers are complementary, not redundant:

- **`.gitattributes` `text=auto` is defense.** The condition becomes harmless: no dirty clone, no aborted switch, no outage.
- **The `line-endings` job is enforcement.** `--renormalize` still detects the drift under `text=auto` (row 3), so the index cannot quietly accumulate CRLF.

`.gitattributes` alone cannot close this, and it is worth being exact about why: **attributes are instructions to a git client, and the commits that cause this never run one.** Dependabot and Copilot write blobs through the GitHub API, so no attribute is consulted at the moment the CRLF enters. Attributes can only change what happens to the blob afterwards. "Committed bytes match the declaration" is inherently a verification step.

### Scope: all 27 rules, not just the CRLF ones

The same experiment against `*.json text eol=lf` dirties the clone and aborts the switch **identically**. That matters, because `package.json`, `package-lock.json` and `*.yml` are what Dependabot edits most often -- the trap was armed on those too and had simply not been tripped yet. So every explicit rule moved to `text=auto`; the default `* text=auto eol=lf` already was.

### Verified a no-op for existing content

`git ls-files --eol` over all **1568** tracked files: the `i/` and `w/` columns are **identical** before and after (only the `attr/` column changes), and `git add --renormalize .` stages nothing except `.gitattributes` itself. Windows still gets CRLF for `.cs`/`.csproj`/`.sln`/`.props`.

### Guarded

`text=auto` reads *weaker* than `text`, so the obvious future "cleanup" is to tighten it back -- which silently re-arms a repo-wide latent outage. A test in `ci-aggregate-workflow.test.js` fails on any bare `text` rule, verified red (`.gitattributes must use 'text=auto', not a bare 'text': *.csproj    text eol=crlf`) and green. 404/404 script tests.

Thanks for the nudge -- this is a better primary fix than the gate alone, and it would not have turned up without asking whether the declaration itself could carry the weight.

Issue #314, opened 2026-07-30
docs: prove a safe EditorWindow capture path and replace draft tooling screenshots


wallstop:
## Problem

PLAN.md WS-7.3 is the last open editor-tooling convergence item. The inspector-overlay screenshot manifest and guide still mark the current images as draft/stale because the Linux-devcontainer-to-Windows-host MCP topology has no proven first-class EditorWindow capture path.

Previous experiments established important safety boundaries:

- Do not call `UnityEditorInternal.InternalEditorUtility.ReadScreenPixel`.
- Do not automate native `PrintWindow` capture.
- Do not switch Unity editor skins programmatically during capture.
- Start from a Personal/light editor, record the actual skin, inspect every image, and clean the host state after capture.
- Scene View and camera capture tools do not prove EditorWindow or Inspector rendering.

## Desired outcome

Establish a repeatable, visually verified capture workflow for the package-owned Inspector, Project Settings, Message Monitor, and Flow Graph surfaces, then replace the draft/stale screenshot set with images of the shipped UI.

## Acceptance criteria

- Document one capture method that works from the configured Windows host editor without blocked native screen-reading primitives.
- Prove the method on the Inspector and at least one package-owned EditorWindow before replacing any manifest entry.
- Capture on Personal/light and record the Unity version and skin with each artifact.
- Visually inspect each image for clipping, stale prototype UI, duplicate windows, incorrect theme classes, and sensitive project data.
- Update `docs/guides/inspector-overlay.md` and the screenshot manifest together.
- Remove every draft/stale marker only when its corresponding image has been replaced and reviewed.
- Leave the host with no temporary scenes, windows, imported prototype assets, or changed skin preferences.
- Keep `scripts/__tests__/design-system-dumps.test.js` green, including the bans on `ReadScreenPixel` and `PrintWindow` automation.
- Run the relevant EditMode editor-tool contract fixtures through Unity MCP after the documentation update.

## Related

- PLAN.md WS-7.3 / D9
- `docs/images/inspector-overlay/README.md`
- `docs/guides/inspector-overlay.md`

wallstop:
Capture-path update from session 177:

I proved a desktop-independent path in Unity 6000.4.6f1 on both acceptance surfaces:

- a real Message Monitor using the shipped view and current data;
- a real Inspector containing `MessagingComponentEditor` and the current Message subscriptions section.

[...]

PR #323, opened 2026-07-31
test(editor): retain a tested offscreen editor surface capture path

wallstop:
## Summary

Closes the mechanism half of PLAN.md WS-7.3 / #314: a retained, test-pinned path that renders a real shipped editor surface into an offscreen render target and writes a cropped 24-bit PNG, never reading the desktop.

Session 177 proved the idea but retained no helper, produced RGBA rather than the manifest's 24-bit RGB, and emitted an IMGUI cursor diagnostic. `Tests/Editor/EditorSurfaceCapture.cs` is that helper, with eight tests in `EditorSurfaceCaptureTests`.

It hosts the surface in a `HideAndDontSave` window, drives the panel through `ValidateLayout()` -> `Repaint(Event)` -> `Render()` into a linear `RenderTexture`, crops to the surface's own `worldBound`, and verifies PNG color type 2 before writing. It restores the render target and `GL.sRGBWrite`, destroys everything it made on both paths, and reports the editor skin instead of switching it.

**No tracked screenshot is replaced.** The artwork half of WS-7.3 stays blocked on the host: it is Pro/dark, the manifest requires Personal/light, and skin switching is banned. #314 stays open for that.

## Findings

Four things were wrong on the first attempt. Each is now fixed in code or recorded in the manifest.

**A never-shown window has no panel**, so there is nothing to lay out or render. That also surfaced a real defect in the shared test host: `EditorWindowTestUtility.CloseWindow` called `EditorWindow.Close()` unconditionally, and `Close()` dereferences a null parent for a window that was never shown. Because the throw came from a `finally`, it replaced the original exception and hid the actual failure. `CloseWindow` now destroys a parentless window instead.

**Omitting the repaint step yields a valid PNG of a blank frame.** `ValidateLayout()` then `Render()` alone produces a correctly sized, correctly typed, entirely empty image. The tests count distinct colors so a blank frame fails rather than passing as a valid PNG.

**A full-canvas readback frames the host window's chrome.** The first visually inspected artifact showed the test host's tab strip with the surface clipped underneath it. Cropping to `worldBound` removes the chrome and is also what gives each image the tight frame the capture list asks for. Render-target rows start at the bottom while UI Toolkit measures from the top, so the vertical origin is `canvasHeight - worldBound.yMax`.

**The `AddCursorRect` diagnostic does not apply to the overlay crops.** Session 177 hit `EditorGUIUtility.AddCursorRect called outside an editor OnGUI` while rendering a whole live Inspector with an IMGUI body. The overlay images the manifest asks for are crops of the package-owned UI Toolkit view, which renders directly and emits nothing.

## Red/green evidence

Live host, Unity 6000.4.6f1, Direct3D11. The capture runs are the red-green record: 3/5 on the first run (teardown NRE), 6/2 once the panel existed (blank frames), 8/0 once the repaint step was added, 8/0 again after the crop change.

- `EditorSurfaceCaptureTests`: 8 passed, 0 failed;
- full `WallstopStudios.DxMessaging.Tests.Editor` assembly: **557 passed, 0 failed** (549 before this session's 8 new tests, so no regression from the `CloseWindow` change);
- visual inspection: `MessageAwareComponentInspectorView` renders to a 720x63 RGB24 crop, 293 distinct colors, no window chrome, no clipping, complete 4-side border;
- Console after capture: one pre-existing unrelated Hot Reload warning, no capture diagnostic, nothing cleared;
- host after capture: seven windows, all standard Unity, no test-host leak, no leaked GameObjects, skin unchanged at `isProSkin=True` / `UserSkin=1`;
- `design-system-dumps.test.js` and `editor-window-test-host.test.js` green, including the blocked-capture-primitive bans;
- full Node/script suite: 406 passed, 0 failed;
- `npm run validate:all`, spelling, prettier, markdownlint, csharpier, pre-commit: passed.

Progress log: `progress/session-180-editor-surface-capture.md`.

Refs #314

---

> [!NOTE]
> **Low Risk**
> Changes are confined to editor test infrastructure and documentation; production runtime code is untouched, with tests guarding render-state cleanup and crop refusal.
> 
> **Overview**
> Adds **`EditorSurfaceCapture`** and nine **`EditorSurfaceCaptureTests`** to close the WS-7.3 **mechanism** half: render real shipped UI Toolkit editor surfaces offscreen into a linear render target, crop to the surface’s `worldBound`, and write manifest-compliant **24-bit RGB** PNGs (color type 2) without desktop or screen capture.
> 
> The capture path shows a hidden test host window, reflects inherited panel `ValidateLayout` → `Repaint` → `Render`, restores `RenderTexture.active` and `GL.sRGBWrite`, and refuses oversized crops instead of silently clipping. **`EditorWindowTestUtility.CloseWindow`** now destroys never-shown windows (no `m_Parent`) so offscreen capture teardown does not NRE.
> 
> Docs in **`docs/images/inspector-overlay/README.md`** document the retained helper; **`progress/session-180-editor-surface-capture.md`** logs findings. **No tracked screenshot PNGs are replaced**—final Personal/light artwork remains blocked on host theme (#314).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2ac96a66689e9394a4e06366bc0d23b03d7abdbc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -393,7 +393,9 @@
     "denoising",
     "labelledby",
     "describedby",
-    "prio"
+    "prio",
+    "ARGB",
+    "IHDR"
   ],
   "ignoreRegExpList": [
     "/[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}/g",
--- a/Tests/Editor/EditorSurfaceCapture.cs
+++ b/Tests/Editor/EditorSurfaceCapture.cs
@@ -1,0 +1,364 @@
+#if UNITY_EDITOR && UNITY_2021_3_OR_NEWER
+namespace DxMessaging.Tests.Editor
+{
+    using System;
+    using System.IO;
+    using System.Reflection;
+    using UnityEditor;
+    using UnityEngine;
+    using UnityEngine.Rendering;
+    using UnityEngine.UIElements;
+    using Object = UnityEngine.Object;
+
+    /// <summary>
+    /// Renders a package-owned editor surface into an offscreen render target and encodes it as
+    /// a 24-bit PNG, without reading the desktop.
+    ///
+    /// The documentation screenshot manifest bans Unity's internal screen-pixel reader, native
+    /// window capture, and programmatic skin switching, because all three read whatever the host
+    /// desktop happens to be showing. This helper never looks outside its own render target: it
+    /// hosts the real shipped view in a hidden window, drives that window's panel through one
+    /// layout and one repaint, and reads back only the temporary target it created. The blocked
+    /// identifiers are listed in scripts/__tests__/design-system-dumps.test.js, which fails on
+    /// any tracked source that names one.
+    ///
+    /// Two details are load-bearing and were established by experiment (see
+    /// docs/images/inspector-overlay/README.md):
+    ///
+    /// - The panel's <c>ValidateLayout</c> and <c>Render</c> are inherited, so they must be
+    ///   reflected with instance, public, and non-public binding flags and WITHOUT
+    ///   <see cref="BindingFlags.DeclaredOnly"/>; declaring-type-only lookup finds nothing.
+    /// - In a linear-color project, a linear render target with <c>GL.sRGBWrite</c> disabled is
+    ///   what matches the colors the real panel shows.
+    /// </summary>
+    internal static class EditorSurfaceCapture
+    {
+        /// <summary>PNG IHDR color type 2: truecolor, no alpha channel. The manifest requires it.</summary>
+        internal const byte PngTruecolorWithoutAlpha = 2;
+
+        /// <summary>
+        /// Byte offset of the IHDR color-type field: 8-byte signature + 4-byte length +
+        /// 4-byte "IHDR" + 4-byte width + 4-byte height + 1-byte bit depth.
+        /// </summary>
+        private const int PngColorTypeOffset = 25;
+
+        private const BindingFlags InheritedInstanceMembers =
+            BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic;
+
+        /// <summary>
+        /// Offscreen rendering needs a real graphics device. CI runs Unity with
+        /// <c>-nographics</c>, so capture tests skip there rather than assert against a device
+        /// that cannot rasterize anything.
+        /// </summary>
+        internal static bool IsSupported =>
+            SystemInfo.graphicsDeviceType != GraphicsDeviceType.Null;
+
+        /// <summary>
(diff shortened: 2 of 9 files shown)


wallstop:
@cursor review
@copilot review

Please focus on the capture helper's failure paths and on whether the tests can pass while the capture is actually broken.

Specifically:

- `EditorSurfaceCapture.Capture` restores `RenderTexture.active` and `GL.sRGBWrite` and destroys the host window plus both textures in a `finally`. Is there a path that leaks one of them, or that leaves a global mutated?
- The crop math converts UI Toolkit's top-left `worldBound` into bottom-up render-target coordinates and clamps to the canvas. Are the clamps correct for a surface that starts inside the canvas but extends past its right or bottom edge?
- The distinct-color assertion is the only thing standing between "rendered the real view" and "wrote a valid PNG of nothing". Is there a cheaper way for it to pass falsely?
- `EditorWindowTestUtility.CloseWindow` now destroys a window whose `m_Parent` is null instead of calling `Close()`. Does that change teardown for any existing fixture that shows its windows?

Evidence in the description: 8/0 capture tests, 557/0 full editor assembly, a visually inspected 720x63 RGB24 crop with no chrome or clipping, a clean Console, and a host left at its seven standard windows.

wallstop:
@cursor review
@copilot review

Both findings are fixed in `9a566c51`.

**Bottom crop clamp keeps full height.** Correct, and the fix goes further than the report: the clamp is gone entirely. `ResolveCropRect` now refuses a surface that does not fit the canvas and names both rects, rather than adjusting the rect to fit. Clamping in a screenshot helper is the wrong shape regardless of which edge it happens on -- it makes the tool built to prevent clipped documentation images capable of silently producing one. `CaptureRefusesASurfaceLargerThanTheCanvas` pins the refusal and also asserts no partial file is left behind.

**Failure restore test skips finally.** Also correct, and it was the more serious of the two, because it meant the `finally` block had no coverage at all. The test now fails from *inside* the try by way of an oversized surface, which throws at the crop step -- past the render-target takeover, past `GL.sRGBWrite`, and past host-window creation. It additionally asserts the host window was closed, which the argument-validation path could never have exercised because no window existed.

Proven by mutation rather than by inspection: deleting the two restore lines from the `finally` now fails **both** restore tests, including the failure-path one (7 passed, 2 failed, `Expected: same as <RenderTexture> But was: null`). Under the previous test body that same mutation went undetected.

Full editor assembly after the fixes: **558 passed, 0 failed**.

Worth re-reviewing with fresh eyes: the refusal predicate now compares `cropX + cropWidth > canvasWidth` and `cropY + cropHeight > canvasHeight` against rounded integers derived from a float `worldBound`. Is there a rounding case where a surface that visually fits is rejected, or vice versa?

Issue #330, opened 2026-07-31
[Feature]: Collapse Multi Merges Onto Main


wallstop:
### Feature Description

Right now something merges, then we trigger perf, then we trigger llm regen.

This should be collapsed into one flow

### Use Case / Motivation

Too many main commit churn

wallstop:
Partially addressed by 66feaae3 (#331) — **leaving this open**, because the largest source of churn is untouched.

The PR body listed this issue under `Closes`, which was an overclaim on my part. (GitHub only applies a closing keyword to the first issue in a comma-separated list, so it did not actually close — but the claim was still wrong.)

**What landed.** Two of the three post-merge auto-commit flows are now one workflow. `post-merge-maintenance.yml` replaces `update-llms-txt.yml` and `update-issue-template-versions.yml`, running both generators in a single job that produces at most **one** `chore: refresh generated repository metadata` commit instead of up to two.

**What remains — and it is the bulk of it.** `perf-numbers.yml` still pushes its own commit:

```
772: git commit -m "docs(perf): auto-update dispatch throughput numbers"
```

That is 10 of the last 25 commits on master. Collapsing it into the same flow is harder than the other two were, and the constraints are worth writing down before anyone tries:

1. **It cannot simply move into `post-merge-maintenance.yml`.** The perf numbers come from a benchmark run on a self-hosted Windows runner holding the organization build lock and a Unity licence seat. The maintenance job is a two-minute `ubuntu-latest` job running two `node` scripts. Merging them would make every llms.txt refresh wait on — and hold — a Unity seat.
2. **The loop break is load-bearing.** `perf-numbers.yml` ignores `docs/architecture/performance.md` and `perf-baseline.csv` on push precisely because its own auto-commit uses a GitHub App token, which re-triggers workflows. Any consolidation has to preserve an equivalent break or it will benchmark itself forever.
3. **The two have different cadences.** Metadata regeneration is cheap and should follow every merge; a benchmark is expensive and arguably should not.

So the realistic shape is probably *deferred aggregation* — perf writes its numbers as an artifact, and the maintenance job commits them alongside the other generated files on the next run — rather than putting the benchmark and the generators in one job. That is a design decision, not a mechanical merge, which is why it is not bundled into a PR about CI signal integrity.

wallstop:
Measured the churn this issue is about, over the last 30 merges to `master`.

[...]

Issue #328, opened 2026-07-31
ci: the stuck-job watchdog reports success while it cannot read runner inventory


wallstop:
## Problem

The stuck-job watchdog has been reporting `success` while unable to do the one thing it exists for. It cannot list runners at all, logs an error, takes no action, and exits 0.

## Evidence

Run [30644904581](https://github.com/Ambiguous-Interactive/DxMessaging/actions/runs/30644904581), job `Audit queued runs and recover dispatcher-stuck ones`, conclusion **success**:

```text
Queued runs older than 300s: 2
gh: Resource not accessible by integration (HTTP 403)
WARN: org-level runner list unavailable (likely 403). Falling back to repo runners.
gh: Resource not accessible by integration (HTTP 403)
ERROR: repo-level runner list also failed; cannot evaluate idle runners. No action issued.
```

Two runs were queued past the threshold and neither was evaluated.

## What it failed to catch

`Performance Numbers` run [30613031270](https://github.com/Ambiguous-Interactive/DxMessaging/actions/runs/30613031270) (`b09a854f`) has been `queued` since **2026-07-31T07:30Z**, with both benchmark legs waiting on `[self-hosted, Windows, RAM-64GB, fast]`. Behind it, the run for `6a70d1f8` has sat `pending` since 09:27Z, blocked by `concurrency: Performance Numbers-refs/heads/master` with `cancel-in-progress: false`.

So the published dispatch-throughput numbers have not refreshed since `2e58122c` at 01:38Z, and the automation meant to notice was inert the whole time. `Self-hosted runner registration preflight` passed for that run at 07:30, which only proves a runner with those labels was *registered*, not online and idle.

## Root cause

Two independent causes, both real.

**1. The watchdog cannot read runner inventory.** It runs with `GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}` and `permissions: actions: write, contents: write`. Both runner endpoints it tries need more than that:

- `GET /orgs/{org}/actions/runners` is org-scoped. `GITHUB_TOKEN` is repository-scoped and can never reach it.
- `GET /repos/{owner}/{repo}/actions/runners` needs `administration: read`, which is not in the `permissions:` block.

Adding `administration: read` alone does not fix it: this repository has **zero repository-level runners** (`GET /repos/.../actions/runners` returns `total_count: 0`); every runner is registered at the organization in group "Default". The repo-level fallback would succeed and report no idle runners, which reads identically to "no runner matches" and still issues no action.

[...]

PR #331, opened 2026-07-31
ci: make the cleanup gate and the stuck-job watchdog mean what they report

### LINKED ISSUE #327: ci: a Unity leg that aborts before lock acquisition still fails Require confirmed Unity cleanup
## Problem

When a licensed Unity leg aborts at its first step, it produces **two** failures: the honest one, and a second from `Require confirmed Unity cleanup` reporting a cleanup that was never needed.

## Evidence

Run [30642224168](https://github.com/Ambiguous-Interactive/DxMessaging/actions/runs/30642224168), job `Unity 6000.3.16f1 playmode`, on PR #326. The leg was scheduled for a superseded head, so `Require current PR head before setup` failed first:

```text
failure    Require current PR head before setup
skipped    ... every step through Require acquired Unity lock ...
skipped    Return Unity license
skipped    Classify Unity cleanup evidence
success    Release organization Unity lock
failure    Require confirmed Unity cleanup
```

Five of the nine legs in that run show the identical shape.

## Why the second failure happens

`Return Unity license` and `Classify Unity cleanup evidence` both carry
`if: ${{ always() && steps.acquire_lock.outputs.acquired == 'true' }}`, so both correctly skip when the lock was never acquired. `Require confirmed Unity cleanup` carries a bare `if: always()` and reads `classification-complete` from the step that just skipped, so it sees an empty value and fails.

In this shape there is nothing to clean: `Acquire organization Unity lock` never ran, so no seat was consumed and no lock was held. The gate is reporting the absence of a classification for work that never started.

## Why it is not simply "add the same condition"

The bare `always()` is load-bearing. If `acquire_lock` itself fails part-way through, `acquired` can read empty while the lock is genuinely held, and that is exactly the leak the gate exists to catch. Gating the gate on `acquired == 'true'` would make the one case it must never miss the case it skips.

So the fix has to distinguish "acquisition never started" from "acquisition failed after taking the lock", rather than collapse both into a skip. Candidates:

1. Have the gate accept an explicit `not-attempted` classification, emitted when `acquire_lock` is `skipped` (as opposed to `failure` or empty output). `steps.acquire_lock.outcome` distinguishes the two, and the aggregate can require `outcome == 'skipped'` before accepting the empty classification.
1. Have `Classify Unity cleanup evidence` run on `always()` and emit `not-attempted` itself, so the gate always receives a value and keeps its bare `always()`.

Option 2 keeps the gate untouched, which is the safer of the two.

## Impact

Cosmetic today: the second failure only ever appears on a leg that is already red, so no verdict changes. It is worth fixing because a spurious cleanup failure is precisely the signal that must stay trustworthy -- the whole four-layer license-return guarantee depends on a reader believing the cleanup gate when it goes red.

After #311 lands (in #326) superseded runs will not schedule these legs at all, which removes the common source. The residual case is a push that lands after a leg has started, where the per-leg head guard fires mid-flight.

## Acceptance criteria

- A leg that aborts before `Acquire organization Unity lock` runs reports exactly one failure.
- A leg where `acquire_lock` fails after taking the lock still fails `Require confirmed Unity cleanup`.
- `Require confirmed Unity cleanup` keeps a bare `if: always()`, or the change that removes it is covered by a case in `scripts/validate-unity-pr-policy.py`'s executed truth table.
- Applies to all five licensed lock windows, not just `unity-tests`.

## Related

- #311 -- removes the common trigger, not the underlying gate behavior.

wallstop:
Makes three pieces of CI automation report what is actually true. Each change
below fixes a case where a check reported a verdict it had not earned.

Closes #327, #328, #330, #333.

## The shape of the bugs

Every defect here is the same shape: **a gate that could not fail**, or a signal
that could not be trusted.

| # | The check said | The truth |
|---|---|---|
| #328 | watchdog green every 5 min | it could not read runner inventory at all; a run sat queued 10 hours |
| #327 | `Require confirmed Unity cleanup` failed | the leg aborted *before* taking the lock, so there was nothing to clean up |
| #333 | `Player build failed` after ~20 min | the host had no `cl.exe`; three licence seats and ~2h burned |
| #330 | post-merge maintenance green | a generator that never converged had its output silently reverted |

## What changed

**`stuck-job-watchdog.yml`** — fails closed on every input it depends on: the
queued-run listing, the organization runner inventory, per-run job listings, and
the cancel-cap state branch. Inventory is read with an organization-scoped reader
App token, because the job's own `GITHUB_TOKEN` can never reach
`/orgs/{org}/actions/runner-groups`. Starvation is now reported from every branch
that can observe it, including the busy and dispatcher-stuck ones — a busy
sibling does not make a starved sibling less starved. An `EXIT` trap guarantees a
step summary even on an unanticipated abort.

**`post-merge-maintenance.yml`** (replaces `update-llms-txt.yml` +
`update-issue-template-versions.yml`) — each generator reverts its own paths when
its checker rejects the output, and the failure flag is reset per regeneration so
it describes the output that actually gets pushed. Conditional steps use
`!cancelled()` rather than `always()`: a generator killed mid-write never reaches
its revert, and `always()` would let the push step commit a half-written file to
`master`.

**`scripts/unity/assert-msvc-toolchain.ps1`** (new) — fails an IL2CPP leg in
seconds, before the organization build lock is taken, when the host cannot
compile. Distinguishes "no Visual Studio" from "toolset present, `cl.exe`
missing" — the latter is what #333 was, and Unity discovers the version directory
and puts it on `PATH` before failing twenty minutes later with a message naming
neither the host nor the cause.

## Three defects the gate shipped with, found in review

Worth calling out because they were invisible to the unit tests:

1. **`exit 1` could not fail a step.** Under this repository's custom `shell:`
   template (`pwsh … -Command ". '{0}'"`) — unlike the built-in `shell: pwsh` —
   no `exit $LASTEXITCODE` is appended. Measured in the workflow's exact shape:
   `exit 1` → **0**, `throw` → **1**. The gate printed its error and passed.
2. **A failing `vswhere` was reported as a missing toolchain.** That blames a
   possibly healthy host and blocks all IL2CPP CI on it.
3. **The argument list was splatted, glob-expanding `-products '*'`.**
   Source-level quoting does *not* prevent this:

   ```
   & $p -products '*'                ->  -products *
   $a=@('-products','*'); & $p @a    ->  -products __pycache__ _site coverage …
   ```

Swept the repository for both classes: all 13 custom-template steps are clean of
`exit N`, and no other PowerShell splat carries a glob metacharacter.

## Testing

`scripts/validate-unity-pr-policy.py` extracts the workflows' `run:` blocks and
**executes** them against stubbed `gh`/`git`, so the assertions are about
behaviour rather than about text matching. ~41 truth-table rows plus standalone
blocks cover the watchdog's verdict space, the post-merge push loop, and the
terminal gate.

`scripts/__tests__/test-msvc-toolchain.ps1` covers every verdict on any platform,
plus four end-to-end guards that drive the script the way the workflow does
against a fake `vswhere` that records its own argv. The fake is described as data
and rendered to `.cmd` or bash per host, because Windows is the only platform the
gate actually runs on.

Findings are pinned by mutation, not by passing: `throw`→`exit 1`, ignored
`vswhere` exit code, dropped path-rooted validation, dropped `-requiresAny`,
re-splatted arguments, lexical version sort, always-true compiler probe, removed
cancel cap, unpersisted cancel state, reader token swapped for `GITHUB_TOKEN`,
ignored inventory-read failure, and tolerated empty group set are all killed.

## Deliberately not done

`visible_to_repository` was **removed** from the runner-group walk rather than
corrected. The REST docs type it only as `"string"` with no format, the endpoint
needs `admin:org` to probe, and both wrong-guess outcomes land in `fail_closed` —
so a wrong guess fails all 288 daily runs. The organization has one group, so the
filter selects what the unfiltered call already returns. Tracked by #335 for the
day a second group appears.

Follow-ups filed: #332, #334, #335, and build-lock#169.

---

> [!NOTE]
> **Medium Risk**
> Changes affect org runner visibility, automatic run cancellation, and default-branch auto-commits, plus a new pre-lock gate on self-hosted Windows IL2CPP legs—high operational impact but limited to CI infrastructure, not shipped product code.
> 
> **Overview**
> CI automation is tightened so **green runs and cleanup gates match what actually happened**, and IL2CPP legs fail before consuming Unity licence seats.
> 
> **Stuck Job Watchdog** now **fails closed** when queued runs, org runner inventory, per-run jobs, or cancel-cap state cannot be read (fixes silent green while a run sat queued for hours). Runner inventory uses an **organization-scoped reader App token** instead of a repo token or empty org fallback. Queued runs are classified with **starvation warnings** even when a sibling job is busy or dispatcher-stuck, and an **EXIT trap** always emits the step summary.
> 
> **Post-merge maintenance** **replaces** the separate `update-llms-txt` and `update-issue-template-versions` workflows with one job that regenerates both artifacts in a **single commit**. Generators use **`!cancelled()`**, revert only their own paths when checks fail, verify reverts, and block push when revert fails; **`llms.txt`** no longer restamps **Last Updated** on byte-identical runs (#330).
> 
> **`Require confirmed Unity cleanup`** in Unity workflows now passes **`acquired: false`** when the lock step was **skipped**, so a pre-lock abort does not double-fail (#327).
> 
> **New `assert-msvc-toolchain.ps1`** runs on **standalone / IL2CPP** matrix legs **before** lock acquisition, distinguishing missing VS, missing toolset, and **toolset without `cl.exe`** (#333). The workflow step uses **`throw`** (not `exit 1`) under the custom `pwsh -Command ". '{0}'"` shell, and **`vswhere`** args are passed without splatting `*`.
> 
> Also: **MSVC assertion tests** in CI, doc/runbook updates, **JS LOC budget** bump, and **cspell** word list reorder.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 100400ff868a9eacd18b6211dbbabe87e30c341b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
@cursor review

wallstop:
bugbot run

wallstop:
### Bugbot round 10 — 1 confirmed, 1 not reproducible

**Confirmed and fixed** — *New scripts missing meta companions* (Low), in `ebf617c2`.
Both new `.ps1` files now carry `.meta` companions with fresh unique GUIDs. Swept
both directions repo-wide: no other tracked file under a `.meta`-using directory
is missing one, and there are zero orphaned metas. The remaining bare files are
dot-prefixed paths Unity does not import, plus two legacy `progress/` entries
that predate `progress/**` being gitignored.

**Not reproducible** — *Exit trap path never exercised* (Medium).
`break_date=True` has a live caller at `scripts/validate-unity-pr-policy.py:1832`,
immediately followed by the `require()` assertions that consume it.

Verified by mutation rather than by reading — deleting `trap on_exit EXIT` from
`.github/workflows/stuck-job-watchdog.yml` and running the validator:

```
validator exit: 1
watchdog unexpected abort: the EXIT trap did not emit 'Watchdog summary'
```

The seam is reachable and load-bearing. Worth noting the finding was *nearly*
right historically: a table-fold refactor did silently delete six assertion
blocks including this one — which you caught — and `1d07e50b` restored them.
The mutation above is what now pins it.

---

Also in this round, self-found on the new MSVC gate and fixed in `239979ae`:

| defect | consequence |
|---|---|
| `exit 1` under the custom `shell:` template | the gate printed its error and **passed** — a no-op check |
| failing `vswhere` reported as a missing toolchain | blames a healthy host, blocking all IL2CPP CI on it |
| **splatted** argument list glob-expanding `-products '*'` | vswhere asked for a product named after whatever sat in the checkout |

The last one is worth flagging generally: source-level quoting does **not** stop
it. Measured on PowerShell 7.6.4 —

```
& $p -products '*'                ->  -products *
$a=@('-products','*'); & $p @a    ->  -products __pycache__ _site coverage ...
```

Swept the repo for both classes: all 13 custom-template steps are clean (none
relies on `exit N`), and no other PowerShell splat carries a glob metacharacter.

wallstop:
### Copilot findings — all three addressed, in `fc7e3041`

**1. `visible_to_repository` (the one that mattered).** You were right to push on
this, and the fix is to stop depending on it. The REST docs type the parameter
only as `"string"` and state no format, so repository *name* and numeric *id*
are not distinguishable from the documentation; the endpoint needs `admin:org`
to probe, which this workspace's token lacks, so it could not be settled
empirically either.

It doesn't need to be. Guess wrong and the call either 4xxs or returns an empty
list — and **both** land in `fail_closed`, so the job fails all 288 of its daily
runs while reporting a credential or outage problem it doesn't have. The
organization registers a single group today, so the filter selects exactly what
the unfiltered call already returns. Dropped: a live failure risk in exchange
for nothing. The comment now records what to do if a second group name ever
appears in the summary — narrow the walk using an id confirmed against a live
response, not against the docs.

**2. Unescaped `case` patterns in `watchdog_gh_stub`.** Confirmed real and fixed.
Needles are literals, so their glob metacharacters are now bracket-quoted.
Measured:

| pattern | vs `runner-groupsXper_page` | vs the real route |
|---|---|---|
| `runner-groups?per_page` | **MATCH** — wrong endpoint answered | MATCH |
| `runner-groups[?]per_page` | NO | MATCH |

So an unstubbed endpoint really could be answered instead of exiting 3, which
defeats exactly the fail-closed property the stub exists to provide.

**3. `{"content": ""}` stubbing a `--jq '.content'` call.** Fixed. The stub does
not implement `--jq`, so routes feeding such a call must supply the
already-extracted value; a JSON object made `base64 -d` fail for stub reasons
rather than modelling an unreadable file. Now `EMPTY_WORKFLOW_CONTENT = ""`.
(The "decode failed" vs "detection said no" distinction stays covered by the
dedicated `NON_DISPATCHABLE_WORKFLOW_BODY` case.)

---

Renaming the summary lines was caught by the harness's own channel assertions,
which is what they're for. Re-verified by mutation that the suite still kills:
dispatch detection forced true, ignored inventory-read failure, tolerated empty
group set, reader token swapped for `GITHUB_TOKEN`, removed cancel cap, and
unpersisted cancel state.

CI was fully green on `475784d5` — 24 checks including script tests on all three
OSes, Unity editmode and playmode, and Bugbot.

wallstop:
bugbot run

wallstop:
### Round 12 — 1 fixed, 3 already fixed (verified by mutation)

**Copilot — `run_watchdog` return annotation.** Fixed in `4bfcefd8`.
`WatchdogOutput` subclasses `str`, so `tuple[int, str]` wasn't *wrong*, but it
hid the `.summary` attribute the step-summary assertions depend on — the
signature described less than the contract. Swept: `run_watchdog` is the only
function that constructs one.

**Bugbot — all three are findings you raised earlier in this PR that are already
fixed.** Each fix is present and load-bearing; deleting it fails the harness:

| finding | fix location | mutation applied | validator |
|---|---|---|---|
| Stuck runs invisible on cap failure | `stuck-job-watchdog.yml:612` | delete the `dispatcher-stuck; queued for cancel` `log_summary` | **exit 1** |
| Sticky regenerate failure across retries | `post-merge-maintenance.yml:248` | delete the per-call `regenerate_failed=0` | **exit 1** |
| Busy verdict hides sibling starvation | `stuck-job-watchdog.yml:578` | delete `report_starvation` from the busy branch | **exit 1** |

All three carry a `(Cursor Bugbot.)` attribution comment from the round that
fixed them. I think the re-report is because the review is against the
cumulative diff versus `master`, where the *original* lines are still visible as
additions — the corrected versions are in the same diff.

No change made for these three. If any of the mutations above looks like the
wrong probe for what you meant, say which and I'll re-test.

---

CI on `fc7e3041`: **33 checks passing, 0 failing** — all nine Unity legs
(2021.3.45f1 / 2022.3.45f1 / 6000.3.16f1 × editmode / playmode / standalone).
The standalone legs passing is the direct confirmation for #333.

wallstop:
bugbot run

wallstop:
### Round 13 — you were right twice; both fixed in `f33dd8c6`

I had refuted three of your findings with mutation evidence. Two of the ones I
had **not** examined turned out to be real, and both are exactly the failure
this PR exists to remove — a check reporting a verdict it never earned.

**1. Partial maintenance commit skips llms (`post-merge-maintenance.yml:165`).**
Confirmed. `Regenerate llms.txt` carried no `if:` at all, so it defaulted to
`success()` and was skipped whenever anything earlier failed. Its sibling runs
under `!cancelled()`, so it still ran, still converged, and the push step still
committed — and the terminal gate counted `skipped` as converged (`success|skipped) ;;`).
Master could take an auto-commit that refreshed the dropdown and never refreshed
`llms.txt`.

Fixed on both sides, because either alone leaves half the hole:
- the generator now carries the same `!cancelled()` as its sibling, so an earlier
  failure makes it **run and fail**, attributed to the generator that failed;
- `skipped` is no longer converged. Both generators now share the gate's own
  condition, so whenever the gate runs they both ran — a skip means something
  unanticipated stopped one.

Worth noting *why* the existing cancellation-policy check missed it: it inspects
only steps that **have** an `if:`, on the sound reasoning that an unconditional
step can't run after a cancel. It was blind to the other half — an unconditional
step also can't run after a **failure**. It now additionally requires every
`gen_*` step to declare `!cancelled()`, discovered by id so a third generator
added later is held to the same rule.

**2. Starvation log denies pending cancel (`stuck-job-watchdog.yml:379`).**
Confirmed. `report_starvation` hard-coded `"Not dispatcher-stuck; no action."`
for every caller — and one of its four callers is the dispatcher-stuck branch,
which logs `"queued for cancel"` on the very next line. Same run id, two
contradictory claims. The disposition is now the caller's to state, since it's
the one thing that function cannot know; the three no-action callers keep it as
the default. The existing `a cancellable sibling does not suppress a co-resident
starvation` fixture now asserts the contradiction is **absent**.

**3. Step summary assertions never used (`validate-unity-pr-policy.py:1026`).**
Not reproducible — `.summary` is read at line 1881:

```python
require(needle in text.summary,
        f"watchdog summary channel: {needle!r} never reached GITHUB_STEP_SUMMARY ...")
```

That assertion fired earlier in this same PR when I renamed a summary line, which
is how I know it is live rather than merely present.

---

Mutations, each against a verified-green baseline (I had one round where the
baseline was already red, which makes every "killed" meaningless — so these were
re-measured): gate accepts `skipped` again · either generator loses
`!cancelled()` · llms revert removed · disposition hard-coded again · stuck
caller drops its disposition — **all killed**.

wallstop:
bugbot run

wallstop:
### Round 14 — Copilot: 1 judgment call recorded, 2 against superseded code

**`cl.exe` presence vs. launch (`assert-msvc-toolchain.ps1:225`).** Fair point,
and deliberate — now documented in the file rather than left implicit, and filed
as #336.

A launch probe (`cl.exe /?`) has real ways to fail on a **healthy** host:
`cl.exe` resolves sibling DLLs (`mspdb*`, `msvcp*`) from its own directory and
the toolchain normally runs under `vcvars`. This gate runs **before** the
organization build lock, so a false failure blocks every IL2CPP leg on that
runner rather than one job — the same shape as the `vswhere` defect fixed
earlier in this PR, where a discovery failure was reported as a missing
toolchain and would have told a healthy administrator to reinstall the workload.

The asymmetry settles it: missing `cl.exe` is what #333 actually was and presence
catches it; corrupt-but-present is rarer and still fails the build, just later.
Doing it properly needs a Windows host where **both** the healthy and the corrupt
case can be demonstrated — #336 lists exactly what that requires — rather than a
change made blind from a Linux container.

**`visible_to_repository` expects a repository ID (`:137`, `:298`).** The
parameter and `repo_name` were both **removed** in `fc7e3041`; neither string
exists in the file any more:

```
$ grep -n 'repo_name\|visible_to_repository=' .github/workflows/stuck-job-watchdog.yml
  (no matches)
```

The walk now enumerates all organization runner groups. Reasoning is in the
commit and in #335 for the day a second group appears.

**The MSVC test uses a `#!/bin/bash` stub and will fail on `windows-latest`
(`ci.yml:778`).** True when first pushed, and CI caught it — fixed in
`475784d5`. The fake is now described as data and rendered per host
(`test-msvc-toolchain.ps1:138` branches to `vswhere.cmd`). Empirically on this
exact tip:

```
pass  Script tests (windows-latest)
pass  Script tests (ubuntu-latest)
pass  Script tests (macos-latest)
```

---

I also stopped trusting the summary line. Enumerating every comment still
anchored to current code (`line != null`) is what surfaced the two real findings
in round 13 that the round-12 summary did not lead with — so this round I worked
from the full list rather than the headline.

CI on `f33dd8c6`: 34 checks passing, 0 failing.

wallstop:
bugbot run

wallstop:
### Round 15 — both new findings real, fixed in `13d844aa`

Enumerating every anchored comment (rather than the summary line) surfaced two I
had not seen. Both were right.

**Bugbot — dead MSVC gate (`unity-benchmarks.yml:215`).** Confirmed. The step is
gated on `matrix.test-mode == 'standalone'`, and that workflow's matrix declares
only `editmode` and `playmode`. It could never run — dead weight implying IL2CPP
coverage on a workflow whose Mono legs never touch `cl.exe`.

This is the PR's own subject matter in miniature: not a check that cannot *fail*,
but one that cannot *run*. So rather than only deleting it, the harness now
asserts the property across every workflow:

| workflow | test-modes | gate | reachable |
|---|---|---|---|
| unity-tests | editmode, playmode, standalone | yes | ✅ |
| perf-numbers | standalone, playmode | yes | ✅ |
| unity-benchmarks | editmode, playmode | **removed** | — |
| release | (no matrix) | no | — |

Mutation-verified both directions: re-adding the dead step is caught, and
dropping `standalone` from a matrix that keeps the step is caught.

**Copilot — `-requiresAny` wording (`assert-msvc-toolchain.ps1:187`).** Confirmed
and corrected. My comment said the switch takes the component "AND" the
workload; it is an any-match switch, so the wording read as the opposite of what
the code does — and could have misled a future edit into tightening a filter
whose looseness is precisely the point (an install pinned to an older toolset
must still pass).

**Copilot — `cl.exe` presence vs. launch.** Standing by the earlier reply:
declined with reasoning recorded in the file, tracked as #336. A bare launch has
real ways to fail on a healthy host, and this gate runs before the organization
build lock, so a false failure blocks every IL2CPP leg on that runner.

---

CI on `9b33a9bd`, checked against the head SHA rather than the PR-level rollup:
**34 success, 4 skipped, 0 failures**, spelling included.

wallstop:
bugbot run

wallstop:
### Round 16 — Bugbot clean; Copilot's flake is real, its fix would have broken the test

**Bugbot on `13d844aa`: ✅ no new issues.**

**Copilot — `FRESH_TIMESTAMP` is wall-clock (`validate-unity-pr-policy.py:1196`).**
The concern is right and now fixed in `59c1bed3`. The timestamp was frozen where
the table is *constructed*, hundreds of lines above the loop that *executes* it
and several standalone blocks earlier, so it ages while the suite runs. Margin
today is wide — ~20s of runtime against the 300s gate — but this file tripled in
size during this PR and CI hosts are slower. It is now a sentinel resolved at
execution time.

**The suggested fix, though, would have removed the flake and the test's point
together.** A fixed *future* timestamp makes the age negative, and a negative age
fails `($now - $created) >= $min` for **every** `$min` — including 0. So the row
would stop detecting the age gate being dropped to 0.

Measured before changing anything, because the comment on that row claims exactly
this property:

| variant | `MIN_QUEUE_AGE_SECONDS` → 0 | age filter deleted |
|---|---|---|
| before (frozen `now`) | caught | caught |
| **after (resolved at execution)** | **caught** | **caught** |
| future timestamp (as suggested) | would **not** be caught | — |

The row needs an age in `[0, MIN_QUEUE_AGE_SECONDS)`. An age of ~0 *at execution*
is the only value that both stays under the gate and keeps the gate's removal
detectable — so the fix was to correct *when* it is evaluated, not *what* it is.

---

That is every anchored finding from both reviewers addressed: fixed, or refuted
with a mutation showing the code is already correct. Full list re-enumerated on
this head rather than read from the summary line.

wallstop:
bugbot run

wallstop:
### Round 17 — real hole, fixed in `74fcefbe`; but `set -e` was not the fix

**Bugbot on `59c1bed3`: ✅ no new issues.**

**Copilot — generator steps omit `set -e`, so a failed revert could let rejected
output be pushed.** The *risk* is real and now closed. The commit step is
deliberately allowed to run after a failed generator, and the only thing making
that safe is the invariant the file itself states: *"the tree can only ever
carry output a checker accepted."* Each generator reverts its own paths to hold
that up — but nothing ever checked the revert worked. A `git checkout` that
fails (pathspec matching nothing, a file the generator deleted) breaks the
invariant silently.

Each generator now **verifies** its revert with `git diff --quiet` and reports
`revert-failed=true`; the push step refuses when either says so.

**`set -euo pipefail` would not have fixed it**, which is worth recording because
it is the obvious-looking fix. Measured both ways:

```
set -uo pipefail   -> step exit=1, file still holds the rejected content
set -euo pipefail  -> step exit=1, file still holds the rejected content
```

Stopping the step earlier does not un-modify a file, and the explicit `exit 1`
already made the step fail either way — so `set -e` changes neither the step's
verdict nor the worktree. Only refusing the push changes the outcome.

Mutations killed: either tree-trust clause dropped from the push condition,
either generator no longer verifying its revert, and a generator no longer
reporting `revert-failed`.

---

That is the second time this round-trip has been productive in the same shape:
the *finding* is sound and the *suggested remedy* would have missed (or, with
the future-dated timestamp last round, actively removed) the thing it was meant
to protect. Both are fixed on the substance rather than the letter.

wallstop:
bugbot run

wallstop:
### Round 18 — Bugbot clean; Copilot's stale-comment catch was right

Fixed in `100400ff`. Removing the dead MSVC step from `unity-benchmarks.yml` left its explanatory comment behind, so the file still described an IL2CPP toolchain assertion it no longer has — and never could, since that matrix is editmode and playmode only.

An incomplete deletion that reads as documentation is worse than the dead step was: the step at least did nothing, while the comment actively tells an operator something untrue. The comment now says *why* there's no gate there and where the real ones live.

Swept for the same class — references to anything this PR removed:

| reference | status |
|---|---|
| the two deleted workflows | only the deliberate "it replaces" line in their successor's header |
| MSVC in `ci.yml` | it *runs* the test suite; not a gate declaration |
| `visible_to_repository` | survives only in the comment explaining why it's gone |

**Bugbot on `74fcefbe`: ✅ no new issues.**

wallstop:
bugbot run

Issue #335, opened 2026-08-01
ci: confirm visible_to_repository's format before narrowing the watchdog's runner-group walk


wallstop:
## Context

`stuck-job-watchdog.yml` enumerates **all** organization runner groups:

```
orgs/${OWNER}/actions/runner-groups?per_page=100
```

It previously filtered with `visible_to_repository=<repo name>`. That was removed
in #331 rather than corrected, because the correct value could not be
established:

- the REST docs type `visible_to_repository` as `"string"` and state no format,
  so repository **name** and numeric repository **id** are both plausible and
  not distinguishable from the documentation;
- the endpoint requires `admin:org`, which no token available in that workspace
  carried, so the two forms could not be compared against a live response;
- **both** wrong-guess outcomes (a 4xx, or an empty list) land in the watchdog's
  `fail_closed`, so guessing wrong fails every one of its 288 daily runs while
  reporting a credential or outage problem that does not exist.

The organization registers a single runner group ("Default") today, so the
filter would select exactly what the unfiltered call already returns. Removing it
cost nothing and removed a live failure risk.

## Why this needs revisiting eventually

Enumerating all groups **over-counts capacity** if a second, restricted group is
ever added. The watchdog cancels runs it believes are dispatcher-stuck, and
over-counting capacity is precisely what could turn a legitimately queued run
into a wrongful cancel. The group names are already logged to the step summary
for this reason.

## Trigger and required work

Act when a second group name appears in the watchdog's `Organization runner
groups:` summary line.

1. With an `admin:org` token, call the endpoint three ways and record the
   responses: unfiltered, `visible_to_repository=<repo name>`, and
   `visible_to_repository=<numeric repo id>`.
2. Narrow the walk using **whichever form the live response confirms**, not
   whichever the docs seem to suggest.
3. Add a fixture row to `validate_stuck_job_watchdog` covering a group that is
   *not* visible to this repository, so the narrowing is pinned.

Until then the current unfiltered walk is correct and safe.

Original finding by GitHub Copilot on #331.

wallstop:
Superseded by #338 — and the premise this issue rested on was wrong.

This issue said the organization registers a single runner group, so dropping the `visible_to_repository` filter cost nothing, and asked that the walk be narrowed only if a second group ever appeared. The first production run of the merged watchdog showed the second group was already there:

```
Organization runner groups: Default, ambiguous-interactive-organization-builds, Default
```

So the trigger condition this issue described had already fired before it was written.

[...]

PR #338, opened 2026-08-01
ci: count only runner groups this repository can actually use

wallstop:
Follow-up to #331, correcting a claim I made in it.

## What was wrong

#331 removed the `visible_to_repository` filter from the watchdog's runner-group
walk, on the stated grounds that *"the organization registers a single group
today, so the filter selects exactly what the unfiltered call already returns."*
The first production run after that merge disproved it:

```
Organization runner groups: Default, ambiguous-interactive-organization-builds, Default
```

Two distinct groups, one repeated by `--paginate`. The premise was simply wrong,
and the risk I filed as hypothetical in #335 was live: runners in a group this
repository may not be allowed to use were being counted as available capacity.

**Why that matters.** Inflated capacity is what turns a legitimately queued run
into a *wrongful cancel* — the audit sees an idle runner carrying the required
labels, concludes the dispatcher is stuck, and cancels a run that was correctly
waiting for a runner it can actually use. No wrongful cancel has occurred: every
post-merge run so far reported `No dispatcher-stuck run found`. The exposure was
real but unrealised.

## The fix

Visibility is resolved through **documented shapes** rather than the ambiguous
query parameter that started all this:

* each group's own `visibility` field — `all` needs no further call;
* `selected` groups are resolved against the group's repositories endpoint and
  compared with `github.repository_id`.

Numeric ids on both sides, nothing to guess. This is what #335 asked for, minus
the `admin:org` probe, because the group object already carries `visibility`.

Groups that exclude us are named in the step summary, and a walk ending with
**no** visible group fails closed — capacity we cannot read is not capacity that
is zero.

Also dedupes the group list by id. The runner dedupe already made the repeat
harmless for the counts, so this is not a correctness fix; it removes a redundant
API call per duplicate and stops the summary printing `Default, …, Default`,
which is exactly the kind of line that makes an operator distrust the report.

## A harness bug that made the first version of these tests worthless

`watchdog_runners` numbered runner ids from 1 on **every** call, and the
inventory is deduplicated by runner id — so two groups' runners collided and the
second group's silently vanished. My multi-group fixture therefore proved
nothing: the case asserting *"an invisible group's runner is excluded"* passed
identically whether or not the exclusion worked. Ids are now distinct per group,
and only then did the mutations start failing as they should.

## Verification

Four mutations, each against a verified-green baseline:

| mutation | result |
|---|---|
| visibility filtering removed (the pre-fix behaviour) | killed |
| an invisible group treated as visible | killed |
| group dedupe removed | killed |
| an empty visible set tolerated instead of failing closed | killed |

New rows: an idle runner in a group we cannot use is not capacity · an idle
runner in a group we can use is capacity · no visible group fails closed · a
group repeated by pagination is walked once.

Refs #335.

---

> [!NOTE]
> **High Risk**
> Changes how the watchdog decides dispatcher-stuck vs legitimately queued runs; over-counting capacity was the path to cancelling runs that were correctly waiting, and this PR is the corrective logic on that automation.
> 
> **Overview**
> The stuck-job watchdog no longer treats **every** organization runner group as capacity for this repo. It filters groups by documented `visibility`, `allows_public_repositories`, and (for `selected` groups) whether `github.repository_id` appears in the group’s repository list—so idle runners in groups this public repo cannot use are not counted and cannot trigger a wrongful cancel.
> 
> Runner groups are **deduped by id** after pagination (fewer redundant API calls and cleaner summaries). Excluded groups are named in the step summary; if **no** group is usable, the audit **fails closed** instead of reporting zero capacity. Inventory logging now says capacity is from groups visible to `${REPO}`.
> 
> `REPO_ID` is passed into the audit step. The unity PR policy harness adds `REPO_ID`, fixes `watchdog_runners` so multi-group fixtures use distinct runner ids, and adds regression cases for invisible groups, public-repo refusal, `private` visibility, fail-closed when nothing is visible, and pagination duplicates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05c2967b80a0435c67c9132a8a7fc29c67b49d66. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/stuck-job-watchdog.yml
+++ b/.github/workflows/stuck-job-watchdog.yml
@@ -376,6 +376,16 @@ jobs:
               runner_group_ids+=("${group_id}")
               continue
             fi
+            # `private` means "every PRIVATE repository in the organization" --
+            # not a selected list. This repository is public, so such a group can
+            # never serve it. Excluded here rather than fetching its repositories
+            # endpoint, which would answer a question that does not apply and
+            # could fail the whole audit closed on an error it did not need to
+            # make. (Cursor Bugbot.)
+            if [[ "${group_visibility}" == "private" ]]; then
+              excluded_groups+=("${group_name} (private repositories only)")
+              continue
+            fi
             if ! gh_reader api --paginate \
                 "orgs/${OWNER}/actions/runner-groups/${group_id}/repositories?per_page=100" \
                 | jq -s '[.[] | (.repositories // [])[] | .id | tostring]' > "${group_repos_json}"; then
--- a/.github/workflows/stuck-job-watchdog.yml
+++ b/.github/workflows/stuck-job-watchdog.yml
@@ -338,7 +338,11 @@ jobs:
           # and `ambiguous-interactive-organization-builds` -- so that premise
           # was simply wrong, and the risk documented in #335 was live rather
           # than hypothetical.
-          if ! group_rows_raw="$(jq -r '.[] | "\(.id)\t\(.visibility // "all")\t\(.name)"' < "${runner_groups_json}")"; then
+          # `!= false`, NOT `// true`: jq's `//` treats an explicit `false` as
+          # absent and yields the right-hand side, so a group that genuinely
+          # refuses public repositories would read as allowing them -- the
+          # exact inversion of the check. Caught by the fixture for it.
+          if ! group_rows_raw="$(jq -r '.[] | "\(.id)\t\(.visibility // "all")\t\(.allows_public_repositories != false)\t\(.name)"' < "${runner_groups_json}")"; then
             fail_closed "could not parse the organization runner-group inventory."
           fi
           group_rows=()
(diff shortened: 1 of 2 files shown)


wallstop:
bugbot run

wallstop:
### All three findings were real; fixed in `3d27513e`

**Bugbot — `allows_public_repositories` ignored.** Confirmed and material: this repository is public (`visibility=public`), so a group that refuses public repositories cannot serve it, and counting its idle runner is exactly the over-count that cancels a live run.

Excluded unconditionally rather than only when we know we're public — that would need another API call and another failure mode, and the cost of being wrong is asymmetric: for a *private* repo this under-counts, and under-counting can only produce a starvation report, never a cancel.

**Writing the fixture for it found my check inverted.** `jq`'s `//` treats an explicit `false` as absent:

```
{"allows_public_repositories": false} | // true   -> true
{"allows_public_repositories": false} | != false  -> false
{}                                    | != false  -> true
```

So `.allows_public_repositories // true` read a group that genuinely refuses public repositories as allowing them — the exact inversion of the check, written in the idiom a reader would assume is correct. The test failed on the first run and that is the only reason it was caught.

**Copilot — `--argjson` can abort without failing closed.** Right. An empty or non-numeric `REPO_ID` makes `--argjson` fail to parse, and under `set -e` that kills the audit *without* reaching `fail_closed` — a job that dies rather than reports. Both sides are compared as strings now.

**Copilot — the `IFS=', '` join.** Right, and measured:

```
a=(one two three); IFS=", "; echo "${a[*]}"   ->  one,two,three
```

Quoted `${arr[*]}` joins with only the FIRST character of IFS. Replaced with an explicit join.

---

Mutations killed, each against a verified-green baseline: public-repo gate removed · `// true` restored · visibility filtering removed · invisible group treated as visible · group dedupe removed · empty visible set tolerated · `--argjson` restored.

wallstop:
bugbot run

wallstop:
### Bugbot round 2 — right again, fixed

**`private` visibility treated as `selected`.** Confirmed. `visibility: private` means *every private repository in the organization*, not a selected list — so only `all` short-circuiting meant a `private` group fell through to the repositories-endpoint branch, which answers a question that doesn't apply to it and could fail the whole audit closed on an error it never needed to make. This repository is public, so such a group can never serve it.

Excluded directly now, same shape as the `allows_public_repositories` refusal and for the same asymmetry: under-counting capacity can only produce a starvation report; over-counting is what cancels a live run.

Mutation killed: `private` treated as `selected` again. Copilot reported no new comments on this head.

PR #343, opened 2026-08-02
feat(editor): summarize multi-object subscriptions

### LINKED ISSUE #297: [Feature]: Summarize subscriptions across a multi-object selection
### Feature Description

The Message subscriptions section added in #295 is hidden when more than one object is selected, because each row describes one component's registrations and showing the first target's rows would misrepresent the rest.

Hiding it is correct but not helpful: selecting several components of the same type is exactly when you want to compare what they are listening to.

### Use Case / Motivation

Selecting every enemy in a scene and seeing at a glance that one of them has a token disabled, or is missing a registration the others have.

### Proposed Solution

Render an aggregate view when `Editor.targets.Length > 1`: one row per distinct (message type, registration kind, priority), with a count of how many selected components carry it and a mixed-state dot when the selected tokens disagree on enabled state. Divergence is the signal worth surfacing.

`MessageAwareComponentSubscriptionsState` is already a plain-C# snapshot, so the aggregation and its tests need no editor panel.

### Alternatives Considered

Keep hiding the section. Acceptable today, but it turns the most useful comparison case into a blind spot.

wallstop:
## Summary

- keep the Message subscriptions section visible for homogeneous multi-object selections
- aggregate registrations by actual message type, registration kind, and priority
- report component coverage plus visible enabled, disabled, or mixed token state
- preserve single-object call counts and token behavior
- document the v3.3.0 behavior and update the existing Unreleased entry

## Why

The inspector previously hid subscriptions whenever more than one object was selected. That avoided misrepresenting the first target, but it also hid the most useful comparison case: finding the enemy or scene component with a missing or disabled registration.

The aggregate counts each selected component once per row, keeps same-name message types distinct internally and through qualified tooltips, does not sum call counts, and removes destroyed targets from the live denominator.

## Validation

- Unity MCP focused fixtures: 67 passed, 0 failed
- Unity MCP full EditMode editor assembly: 576 passed, 0 failed, twice back-to-back
- full Node.js script suite: 406 passed, 0 failed
- `npm run validate:all`
- docs snippet compilation
- CSharpier, Prettier, markdownlint, spelling, line-ending validation, and `git diff --check`
- two adversarial review loops, final result: zero findings

Closes #297

---

> [!NOTE]
> **Low Risk**
> Editor-only inspector aggregation over existing token metadata; no runtime messaging API or dispatch changes.
> 
> **Overview**
> **Homogeneous multi-select** on `MessageAwareComponent` inspectors no longer hides the **Message subscriptions** section. The fallback editor collects every selected component and feeds them into the existing subscriptions UI instead of omitting the block when `Editor.targets` has more than one entry.
> 
> For **multiple targets**, registrations are **aggregated** by actual message `Type`, registration kind, and priority. Each row shows how many selected components carry that pattern (`n of m selected`), with a status dot for **all enabled**, **all disabled**, or **mixed** token state (new amber `.dx-sub__mixed` style plus tooltips). Aggregate rows skip summed call counts so per-component differences stay visible; **single-selection** behavior (call counts, listening/disabled header) is unchanged.
> 
> **CHANGELOG** and the inspector overlay guide document the v3.3.0 multi-object behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 59a2524e791078185eb7c08bfcd316786ca14a76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Editor/CustomEditors/MessageAwareComponentFallbackEditor.cs
```diff
@@ -149,22 +150,44 @@ Func<MessageAwareComponentInspectorState> resolveState
...
+            if (targets == null || targets.Length == 0)
+            {
+                return Array.Empty<MessageAwareComponent>();
+            }
+
+            MessageAwareComponent[] components = new MessageAwareComponent[targets.Length];
+            int componentCount = 0;
+            for (int index = 0; index < targets.Length; index++)
+            {
+                if (targets[index] is MessageAwareComponent component)
```
wallstop: Should we also null-check?

Editor/CustomEditors/MessageAwareComponentFallbackEditor.cs
```diff
@@ -149,22 +150,44 @@ Func<MessageAwareComponentInspectorState> resolveState
...
+        private static MessageAwareComponent[] ResolveComponents(Editor editor)
         {
             UnityEngine.Object[] targets = editor.targets;
-            return targets is { Length: 1 } ? targets[0] as MessageAwareComponent : null;
+            if (targets == null || targets.Length == 0)
+            {
+                return Array.Empty<MessageAwareComponent>();
+            }
+
+            MessageAwareComponent[] components = new MessageAwareComponent[targets.Length];
```
wallstop: Why don't we just do the simple thing and use a list?

wallstop:
Current head `e5d25aef` is ready for automated and human review.

@cursor review

@copilot review

@wallstop Human review requested for the multi-object subscription aggregation, Unity object lifetime handling, and single-selection compatibility.

wallstop:
New head `59a2524e` addresses both current human review threads: selected targets now use a simple list and explicitly skip Unity-null objects. Focused Unity MCP verification is 67 passed / 0 failed.

@cursor review

@copilot review

@wallstop Human review requested on the exact current head.

Editor/CustomEditors/MessageAwareComponentFallbackEditor.cs
wallstop: Fixed in `59a2524e`. The resolver now assigns each selected target to a local and explicitly skips `target == null`, which also covers Unity's destroyed-object null semantics.

Editor/CustomEditors/MessageAwareComponentFallbackEditor.cs
wallstop: Fixed in `59a2524e`. The resolver now builds a capacity-sized `List<MessageAwareComponent>` directly and returns it as `IReadOnlyList<MessageAwareComponent>`, removing the temporary array and copy.

wallstop:
bugbot run

Please review exact current head `59a2524e791078185eb7c08bfcd316786ca14a76`; the earlier tagged request has not produced a current-head Bugbot result.

PR #352, opened 2026-08-04
Make the Message Monitor readable at a glance

wallstop:
Closes #344. Also closes the tracked half of #349.

## Why

The reporter's screenshot is the **snapshot** Message Monitor in the Diagnostics
Tooling Exerciser scene. Reading the source against it pinned every complaint in
the issue to a concrete cause:

| Reported | Cause |
| --- | --- |
| Stack trace always shown, should collapse | `CreateRow` appended the whole stack trace to every row card |
| Unclear collapsible toggles at the top | The two lane `ScrollView`s were crushed to a single line plus scroll arrows |
| Text rendered off screen at the bottom | The root was a fixed stack; nothing could shrink, so the details pane and component panel left the window |
| Colors unexplained; red reads as a problem | Snapshot mode had no legend and no taxonomy chips at all; live mode only had letter chips |
| Lanes should be filter buttons | A lane was three wrapping paragraphs next to a separate `Filter` button |
| Unclear whether live or deduped | Nothing on either surface named the mode or said what one row stands for |

## What changed

The snapshot Monitor is rebuilt on the same design-system surface as the live log:

- **Rows.** Fixed-height `.dx-row` columns carrying the route kind, message type,
  context, and dispatch id, under a `.dx-list-header`. No stack trace in a row.
- **Details.** The design system's `.dx-detail` with an `EMISSION` card. The stack
  trace is a disclosure that starts closed. Same change applied to the live pane.
- **Legend and filters are the same control.** Three named, counted taxonomy chips
  (`Untargeted 3`, `Targeted 9`, `Broadcast 6`), each drawn in the color that marks
  its kind in every row. Clicking one hides or shows that kind, and **Copy JSON**
  follows them. The live view's `U`/`T`/`B` chips are now named too.
- **Breakdown.** The two lane panels became one collapsed disclosure of clickable
  filter pills. Counts and member lists moved into tooltips.
- **Component Diagnostics** moved behind its own collapsed disclosure.
- **Mode is explicit.** `SNAPSHOT` and `LIVE` badges plus a one-line statement of
  what a row stands for, so the merged `N` column is no longer a mystery.
- **Layout.** The log flexes; the details pane holds its height.

`BuildMonitorUi` now renders from a single state object. The old
`onFilterChanged` + `onSelectedEntryChanged` + `applyFilterText` +
`RefreshCurrentSnapshotContent` dual path collapsed into one
`Action<MessageMonitorViewState>` host notification, so the surface behaves the
same with or without a host.

[...]

Issue #349, opened 2026-08-03
[Bug]: the "progress" directory should not be checked in


wallstop:
### Description

It is machine-local. It needs to be git-ignored and purged from history

### Steps to Reproduce

The "progress" directory exists on latest master.

### Expected Behavior

It doesn't exist

### Actual Behavior

It does exist.

### Unity Version

2021.3 LTS

### Package Version

3.2.1

### Platform

- [ ] Editor
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] iOS
- [ ] Android
- [ ] WebGL
- [ ] Other

wallstop:
Half of this is done in #352.

`progress/**` has been in `.gitignore` for a while, but a gitignore entry does not untrack files that are already in the index, which is why 24 session logs were still on `master`. #352 drops them from the index, so after it merges the directory is gone from the tree and future session logs stay machine-local.

The other half - purging them from history - is a rewrite and force-push of `master`. That invalidates every open branch and every local clone, so I have not done it unilaterally. Say the word and I will prepare it (`git filter-repo --path progress --invert-paths`), coordinate it against the open branches, and re-run the branch-protection checks afterwards.

Leaving this issue open until that decision is made.

wallstop:
Current state, checked on `f81216d7`:

- `git ls-files progress` returns nothing. The directory is fully untracked.
- `.gitignore:391` has `progress/**`, and `:400` has `progress.meta`.
- The last commit that touched `progress/` on `master` is `924700c7`, 2026-08-11.

So "it needs to be git-ignored" is done, and stays done - a new session record cannot be committed by accident.

"purged from history" is not, and cannot be from a pull request. 68 commits still carry those files, and rewriting them means a force-push of the default branch: every open branch and pull request has to rebase, every commit SHA quoted in an issue, a changelog entry, or a progress record stops resolving, and the pinned base SHAs in this repository's own issue history go with them. That is an owner decision and an owner action.

Worth deciding explicitly rather than leaving open indefinitely: the tracked-state half is complete, and the history half is a one-time operation whose cost does not go down by waiting.

From https://github.com/Ambiguous-Interactive/DxMessaging/pull/438.

wallstop:
What I mean is deleting them/removing them from the repo - they should stop existing on latest master.

wallstop:
## Half of this is already done

`progress/` is ignored and nothing under it is tracked, as of this session:

```text
[...]

PR #355, opened 2026-08-04
Make Message Monitor rows act, not just read

### LINKED ISSUE #344: [Bug]: Tool UI (Message Monitor) Improvements
### Description

The Message Monitor has a few issues:

Image

- There is too much data in each message. The stack trace is always displayed, I think it needs to be collapsible (and default collapsed)
- There are a lot of collapsible toggles at the top that aren't clear of their purpose
- In the current view, there is stuff rendered off screen (lots of text at the bottom), I don't know what that's about.
- We should clarify what the colors mean at the top (and have really easy, nice filters for each color). Right now the red (Targeted) seems like it might indicate a Problem
- All of the lanes and stuff should, I think, be filterable buttons - right now there's just a lot of "info" on the screen
- It is unclear if the messages are live, or deduped

### Steps to Reproduce

Run the message monitor in the Diagnostic Tooling Exercise scene

### Expected Behavior

A better UX

### Actual Behavior

Less good UX

### Unity Version

6000.x

### Package Version

3.2.1

### Platform

- [x] Editor
- [x] Windows
- [ ] macOS
- [ ] Linux
- [ ] iOS
- [ ] Android
- [ ] WebGL
- [ ] Other

### Additional Context

_No response_

wallstop:
Answers the second round of owner feedback on #344. Session 190 made the Message Monitor
readable; this makes it actable.

## What the screenshot settled

Two of the seven items were misdiagnosed by the text alone.

**"We need padding/margins on the left of these items."** The attached screenshot is
**snapshot** mode, and the circled area is the log's ROUTE column at the window edge. The 14px
gutter lived on `.dx-col-time` / `.dx-row__time`, and snapshot mode renders no time column. The
gutter moves onto `.dx-row` and `.dx-list-header`, so whichever column comes first inherits it in
either mode.

**"We can't go 'Un-Live' once live."** A `Snapshot` button already existed - at the far end of a
**wrapping** toolbar. This was discoverability, not a missing control. Both modes now switch from
the badge that names the mode (`SNAPSHOT` / `LIVE`), and live mode's explicit button moves up
beside it.

## The rest

| # | Ask | Answer |
| --- | --- | --- |
| 2 | Link/click the contexts | The context selects and pings its object in the Hierarchy while that object is alive, and stays readable but inert once it is gone |
| 3 | Component Diagnostics is not resizable | Component Diagnostics and the stack trace carry drag handles; the handle raises the `max-height` as it drags, because a cap left in place silently wins over the new height |
| 4 | "Type" should take to source | The message type carries an **Open source** button when its declaring file resolves |
| 5 | Stack trace has "extract stack trace" rows, should link to source | One row per frame, each with its own **Open** button; Unity's own capture frames are dropped so the first row is the emitting call site |
| 7 | No pointer intelligence for clickable things | Every interactive shape declares `cursor: link`, guarded by a parameterized invariant |

## Structure

The Flow Graph owned the package's only source-resolution machinery (~1000 lines: a lazy
per-assembly source index with retry/merge/backoff, a declaration parser, a stack-frame location
parser, the link button). Copying it into the Monitor would leave two copies of subtle logic, so
it moved to `Editor/DxMessagingEditorSourceLinks.cs` and both windows call it.
`FlowGraphSourceLocation` became `SourceLocation`, the link button's name lost its `flow-graph`
segment, and **no forwarders were left behind** - the Flow Graph and its 60 test references point
at the new home directly. The move is behavior-preserving by construction and measured: the
focused Flow Graph fixture is 175/0 before and after.

New shared seams: `ReadCallSiteFrames` (one definition of "a frame worth showing"),
`MakeActivatable` (click + Return/Space + focusable + cursor, one definition of "interactive",
replacing three hand-rolled copies), `CreateSourceLookupName` (assembly-qualified name to the
`Ns.Type [Assembly]` shape the resolver reads, splitting after the last `]` so constructed
generics survive), `FindContextObject` / `TryRevealContext`, and
`DxMessagingEditorTheme.CreateResizeHandle`.

`MessageMonitorEntry` gained `ContextInstanceId`. The id is stored rather than the object,
because a log outlives the scene it recorded and an id that fails to resolve is exactly the
"this row can no longer be opened" answer the detail pane needs.

## Measuring caught two defects reasoning would not

- The new row gutter pushed the fixed columns past the edge at the 360x260 layout case
  (`dx-col-count` at xMax 373 in a 360px window). The fixed columns now shrink; route and context
  already ellipsize.
- Two existing tests fed a stack trace that was **only** a capture frame, so filtering left
  nothing. Their fixtures now carry what `MessageEmissionData` actually captures - capture frames
  on top, real frames beneath - which is better coverage than they had.

## Guards

- `EveryInteractiveStylesheetRuleDeclaresACursor`, parameterized over the nine interactive shapes,
  so a new one that forgets the cursor fails rather than shipping the same gap again.
- A measured gutter contract on the real laid-out window, for both the header and a row.
- A parameterized live/destroyed context-link contract.
- A resize contract proving a drag past the shipped cap takes effect.
- A data-driven assembly-qualified-name conversion including a constructed generic.

## Also here

- **Incorporates Dependabot #351** (seven `ambiguous-organization-build-lock` action bumps onto
  `e6fa595`), byte-identical to its branch, so this session lands one PR rather than two. It also
  carries the piece Dependabot could not: `scripts/__tests__/ci-aggregate-workflow.test.js`
  requires the copyable acquire examples in `docs/ops/ci-and-github-settings.md` and
  `docs/ops/ambiguous-release-migration.md` to be pinned at the same SHA the workflows use, and
  Dependabot only rewrites `.github/`. #351 can be closed in favour of this.
- **#349 needs no code here.** #352 already untracked `progress/`, and `.gitignore` already
  carries `progress/**`. What remains is the history purge, which rewrites every published SHA
  and needs an explicit owner decision.

## Local verification

| Check | Result |
| --- | --- |
| Full Editor + Allocations assemblies (host editor, 6000.4) | **896/0** |
| Focused Monitor + LiveView + Theme fixtures | 152/0 |
| Flow Graph fixture, after the extraction | 175/0 |
| Docs snippets (`dotnet test .docs-tests`) | 441/0 |
| `npm test`, `npm run validate:all`, spelling, prettier, markdownlint, actionlint | green |

## Note on master CI at `050a2541`

Not a regression from #352, and not addressed by code here. `Unity 2021.3.45f1 editmode` died
4 minutes in on DAD-MACHINE: it acquired the org lock and then no step after it ran - not the
test runner, not the `if: always()` license return, not the lock release, not the post steps -
and its log blob was never uploaded. That is a runner abort, not a test failure. Every leg
ELI-MACHINE ran on the same commit is green, including 2021.3 playmode, 2021.3 standalone and
2022.3 editmode. The perf leg failed separately with `System.OutOfMemoryException` inside IL2CPP
codegen, against an Editor-only change, on a workflow whose previous 11 master runs were green.
Both want a re-run rather than a fix.

Closes #344.

---

> [!NOTE]
> **Medium Risk**
> Large editor extraction plus a CI artifact path change (cache miss until repopulated); workflow SHA bumps affect licensed Unity lock/cleanup on every matrix leg.
> 
> **Overview**
> Closes **#344** by making the Message Monitor (live and snapshot) **act on** log rows, not only display them.
> 
> **Diagnostics UX:** Message types get **Open source** when declaration resolution succeeds; live **context** rows select/ping the Hierarchy object while it still exists; stack traces render **one frame per row** with per-frame **Open** (Unity capture frames stripped). The **LIVE** badge and a **Snapshot** control beside it exit live mode; list **left gutter** moves to rows/headers so snapshot mode is not flush to the edge; **Component Diagnostics** and stack sections use **drag resize** via `DxMessagingEditorTheme.CreateResizeHandle`; interactive controls get **`cursor: link`** (including toggle innards where USS does not inherit).
> 
> **Refactor:** ~1k lines of source-indexing, stack parsing, and link buttons move from `DxMessagingFlowGraphWindow` into **`DxMessagingEditorSourceLinks`**; Flow Graph calls the shared API with no forwarders. Monitor entries store **`ContextInstanceId`** for safe linking after objects are destroyed.
> 
> **CI / infra:** Generated UPM host projects default to **`.artifacts/u/<version>-<mode>/`** (shorter path for Windows **MAX_PATH** / deep PackageCache); workflows, caches, and LLM docs follow. **`ambiguous-organization-build-lock`** composite actions pin to **`e6fa595`** across Unity workflows (incorporates Dependabot #351).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 360b8e042616109ed1f839c09d8c3812e563666f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


wallstop:
bugbot run

cc @copilot - please review. Focus areas: the ~1000-line source-link extraction in `93f94c7f` (verify nothing was lost or subtly changed versus `master`), `CreateSourceLookupName` on constructed generics, `MakeActivatable` StopPropagation versus the row-click handler, `CreateResizeHandle` pointer capture, and Unity 2021.3 floor validity for `cursor: link` / `cursor: split-resize-up-down` and the pointer-capture APIs.

wallstop:
### Bugbot: both findings confirmed and fixed

Both were real, and an independent adversarial pass on this branch had reached the same two
conclusions, so they land together in the next push.

**1. "Source link never appears after index" (Medium)** -- correct, and the more serious of the
two: `TryResolveMessageSource` is lazy, so the first lookup queues a background index and
answers "not yet". The Flow Graph re-renders on `MessageSourceIndexChanged`; the Monitor never
subscribed, so on the path a reader actually takes -- fresh domain reload, open Monitor, select
a row -- the Open-source button never appeared at all. Ask (4) was effectively not delivered.

The window now subscribes in `OnEnable` and unsubscribes in `OnDisable`. The handler is
deliberately **narrower than the Flow Graph's**: in snapshot mode it re-renders only the detail
pane via a new `TryRerenderDetails` seam, because a full refresh would rebuild the log and take
the reader's scroll position with it -- the same loss a selection change already avoids. In live
mode it calls `RenderLiveBody()`, which also clears the detail pane's memoization; you were
right that the memoized path would otherwise keep showing the linkless pane for as long as the
row stayed selected.

**2. "Resize handle ignores mouse button" (Low)** -- correct. `PointerDownEvent` now returns
early unless `evt.button == 0`. While there, the drag origin reads `style.height` rather than
`resolvedStyle.height`, so successive drags no longer jump from a different origin than the last
one ended at.

### Also fixed in the same push, from the adversarial pass

- **A dragged height was discarded on the next content rebuild.** `RenderContent` runs on every
  filter keystroke, so the resize survived until the next character typed. `MonitorUi` now
  carries the dragged heights, exactly as it already carries the disclosure-expanded flags.
- **`cursor` is not inherited in USS** (confirmed against the 2021.3 property reference). A rule
  on `.dx-row` alone painted the pointer over its 14px gutter and nothing else, because the
  column labels cover the rest of the row. Declared on the row children, the route dot, and the
  toggles' inner parts. The invariant test now asserts the exact expected cursor rather than
  "some cursor", which `cursor: arrow` would have satisfied.
- **`.dx-resizer` was `flex-shrink: 0` and sits outside any scroll view**, so in a panel squeezed
  below its own floor it would lay out past the window edge -- the exact class
  `BuildMonitorUiKeepsNonScrollingSectionsInsideTheWindow` guards, and a plausible 2021.3-only
  failure. It shrinks now.
- **`ApplyResizedHeight` set a flex basis, not a height**, on a `flex-shrink: 1` target, so a
  drag past the available space was a no-op on screen. It clears `flex-shrink` too.
- **Saying "not captured" about a trace that WAS captured** but held only Unity capture frames.
  Those two states now read differently.
- **Test honesty**: the resize test called the apply helper directly and read the value back --
  it passed with every callback in `CreateResizeHandle` deleted. It now drives real
  pointer-down/move/up through the handle. The context test's "destroyed" case passed id `0`,
  which short-circuits before the object lookup and only proved "no context captured"; it now
  destroys a real object. Added the missing positive source-link test (there was only a negative
  one, which is why finding 1 went unnoticed) and a live-badge exit test.
- Plus: a `Q<T>(null)` that would have matched the key label, the shared drain still logging
  "DxMessaging Flow Graph", three usings the extraction orphaned, and the file/namespace
  mismatch (`DxMessagingEditorSourceLinks.cs` moved into `Editor/Windows/`).

### CI note

`Perf benchmarks 6000.3.16f1 standalone` red here is **not this change**. Its job on
`ELI-MACHINE` ran five minutes entirely inside setup: `Run Unity Test Runner` never started, none
of the `if: always()` cleanup steps ran, and its log blob returns `BlobNotFound`. That is
byte-for-byte the signature of the `master` failure filed as #356, now reproduced on the second
Windows runner. This PR changes editor C#, USS, docs and workflow action pins -- nothing an
IL2CPP player build reads.

`copilot` / `copilot-pull-request-reviewer` red is the external account's quota, as the bot
itself says above, and neither is a required check.

wallstop:
bugbot run

Both prior findings are addressed in `2034d8fd`, along with six more from an adversarial pass. Please re-review at this head; the highest-risk areas are the narrowed `HandleMessageSourceIndexChanged` (detail-pane-only in snapshot mode, full body re-render in live), `ApplyResizedHeight` clearing `flex-shrink`, and `.dx-resizer` becoming shrinkable so it cannot lay out past the window edge in a squeezed panel.

wallstop:
Correction to my CI note above. I attributed the perf red to host resource exhaustion; comparing three occurrences side by side rules that out.

All three aborts -- across **two runners, two workflows and three commits** -- died at the *same step boundary*: after step 18 `Upload Unity editor validation diagnostics` and before step 19 `Run Unity Test Runner`. One of them died 2m44s in, before any build work at all. Random OOM does not land on the same boundary three times out of three.

Detail and the revised lead are in #356. The conclusion for this PR is unchanged and, if anything, firmer: the perf red is not this change, which touches editor C#, USS, docs and workflow action pins only. But it does mean `Unity CI Success` cannot be reached while this persists -- it is a runner-side blocker, not something a commit here can clear.

wallstop:
bugbot run

Three of the four latest findings were real and are fixed in `360b8e04`; each was verified before changing anything.

- **Resize drag starts from zero** — confirmed by measurement on the host editor: an UNSET height reports `StyleKeyword.Null` with `value = 0`, a SET one reports `Undefined` with the real value. My check was inverted, so the first drag started from 0. Fixed.
- **MAX_PATH guard breaks on PS 5.1** — correct, and `bootstrap-windows-runner.ps1:224` already documents avoiding `$IsWindows` for exactly this reason. Now uses `DirectorySeparatorChar`. CI runs pwsh 7, which is why it stayed hidden.
- **Keyboard activation fires twice** — correct, and self-inflicted: I had added the character branch on a diagnosis I later retracted. Reverted to the key code alone, and the test now asserts the paired character event must NOT activate again.

**Rejected — MAX_PATH length misses path separator.** The 171 constant already includes the leading separator. Measured against the real failing path: the project path is 96 characters, the remainder begins with `\`, and 96 + 171 = 267, exactly the observed length. There is no off-by-one; adding one would make the guard reject a project that fits.

Please re-review at this head.

PR #405, opened 2026-08-14
Stop a superseded Unity run from holding the concurrency group

### LINKED ISSUE #332: ci: a Unity run superseded mid-flight blocks the next run behind legs that only abort
## Problem

A Unity Tests run that is superseded **mid-flight** keeps holding its concurrency group, and the run for the commit that actually matters cannot start until every one of the stale run's remaining legs has been handed a self-hosted runner purely so it can abort on the head guard.

#311 removed the case where a run is superseded *before* it starts (the `head-check` job skips the matrix). This is the residual case #327 names in passing: "a push that lands after a leg has started, where the per-leg head guard fires mid-flight."

## Evidence

PR #331, three pushes in close succession.

| run | head | state |
| --- | --- | --- |
| [30660955492](https://github.com/Ambiguous-Interactive/DxMessaging/actions/runs/30660955492) | `bc3764fc` | **queued** 50+ minutes, 2 legs still `queued` |
| [30663226224](https://github.com/Ambiguous-Interactive/DxMessaging/actions/runs/30663226224) | `e40abaee` | cancelled |
| [30663681771](https://github.com/Ambiguous-Interactive/DxMessaging/actions/runs/30663681771) | `967df6ae` (current) | **pending**, zero jobs created |

`unity-tests.yml` declares:

```yaml
concurrency:
  group: unity-tests-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: false
```

`cancel-in-progress: false` is deliberate and correct -- it exists so an in-flight benchmark or a leg holding the organization Unity lock is never aborted mid-lease. But it also means run 3 sat `pending` with **no jobs at all** while run 1 waited on runners for legs whose only remaining act was to fail `Require current PR head before setup`.

In run 1, every already-scheduled leg had done exactly that:

```text
failure    Require current PR head before setup
skipped    ... everything through Require acquired Unity lock ...
success    Require confirmed Unity cleanup
```

So the work being queued for was guaranteed to be discarded, while the head that needed validating could not be scheduled. Cancelling run 1 by hand moved run 3 from `pending` to `queued` immediately.

## Why it is worth fixing

- **It serializes badly with `max-parallel: 1`.** Each doomed leg waits its turn for a scarce self-hosted Windows runner before aborting in seconds.
- **It makes a PR look green when it is not.** `gh pr checks` reports no Unity entries at all for a run in `pending` (no check runs exist yet), so a reader -- human or automation -- sees only the passing checks and can conclude the PR settled. That is how this was nearly missed here.
- **It burns the same scarce runners** the `fast`-labelled benchmark legs compete for, which is the resource #328 and #298 are both about.

## Constraints on any fix

- Do **not** flip `cancel-in-progress` to `true`. A leg holding the organization Unity lock must not be cancelled mid-lease; that is the seat-leak the four-layer license-return guarantee exists to prevent, and #83 shows what happens when a lease is disturbed.
- The fix must distinguish "leg has acquired the lock" from "leg is still queued and has acquired nothing." Only the latter is safe to drop.

## Candidate approaches

1. **Let the head guard cancel the run rather than fail the leg.** When `Require current PR head before setup` fires on a leg that has not acquired the lock, call `gh run cancel` on its own run id. The already-running licensed legs finish their lease normally; the queued ones vanish, and the group frees as soon as the last live leg completes.
1. **Split the concurrency group by phase**, so unlicensed setup legs and licensed legs do not share one group. More invasive, and risks weakening the serialization that protects the lock.
1. **A pre-leg freshness gate at the job level** (`if:` on the matrix job reading a fresh live-head lookup) so a superseded leg is *skipped* rather than *scheduled and failed*. A skipped job needs no runner, which removes the queueing entirely. This looks like the smallest change with the largest effect, and it composes with #311's existing `head-check` output.

Option 3 first; option 1 as the fallback if a job-level `if:` cannot see a fresh enough head.

## Acceptance criteria

- A run superseded mid-flight does not make the next run sit in `pending` with zero jobs.
- A leg that has acquired the organization Unity lock is never cancelled or skipped by this mechanism.
- A superseded leg that has not acquired the lock consumes no self-hosted runner time.

## Related

- #311 -- the same shape, for runs superseded *before* they start.
- #327 -- names this residual case; its own fix (exactly one failure per aborted leg) is what makes these runs cheap to identify.
- #298 -- self-hosted runner contention generally.

wallstop:
## Problem

A `Unity Tests` or `Performance Numbers` run that is superseded **mid-flight** keeps its
concurrency group. `cancel-in-progress: false` is deliberate -- a hard cancel mid-lease is the
seat leak the four-layer license-return guarantee exists to prevent -- so the superseded run
holds the group until every remaining leg has been handed a scarce self-hosted Windows runner
purely so it can abort on `Require current PR head before setup`. The run for the commit that
matters sits `pending` with **zero jobs**, which is also why `gh pr checks` reports no Unity
entries at all and the pull request reads as settled (#332).

## What the measurement says

Every failed leg of the last 60 runs of each workflow, joined against its own step outcomes:

| | |
| --- | --- |
| Doomed legs that consumed a self-hosted runner | 41 |
| Their total on-runner time | 8.0 min |
| Their total queue wait for a runner | 440.6 min |

The wait is the entire cost -- 55x the work. That rules out the fix this issue first proposed
(have the head guard cancel its own run). Within one superseded run the doomed legs run
back-to-back once a runner frees up: run `31753050140` shows queue waits of 1624s, 1637s and
1650s, so self-retiring on the first leg would have saved **26 seconds**. Run `31759789857`
(PR #402, yesterday) is the shape that matters: **one** doomed leg, 1991s of queue wait, 11s of
work. Self-retirement would have saved nothing at all there.

What costs 33 minutes is not the doomed legs. It is queueing the current head behind them.

## Fix

Keep `cancel-in-progress: false`. Move the head into the **group identity** instead:

```yaml
group: unity-tests-${{ github.event.pull_request.number && format('{0}-{1}', github.event.pull_request.number, github.event.pull_request.head.sha) || github.ref }}
```

- **Push runs are unchanged.** `github.event.pull_request` is null, so the key is still
  `github.ref` and consecutive default-branch pushes stay serialized. That is the property the
  perf workflow's group comment asks for, and it is preserved exactly.
- **Pull-request runs partition per head**, so a superseded run cannot hold the group.
- **Nothing is cancelled**, so the constraint that a leg holding the organization lock must
  never be cancelled mid-lease is satisfied by construction rather than by a check.

Two runs of one pull request can never both do licensed work: each leg passes
`Require current PR head before setup`, then `Require current PR head before lock acquisition`,
and `acquire-build-lock` itself takes the pull-request identity and drops stale waiters. The
mutual exclusion that matters was always the organization build lock, never this group.

Job-level preflight groups take the same key. Theirs was unreachable while the workflow group
serialized every run of a pull request. Left on `github.ref` it would now let a newer run cancel
a superseded run's preflight, skip that run's whole licensed matrix, and leave its aggregate
reporting a result shape `Verify Unity CI result shape` does not enumerate.

## Scope of the class

Only two workflows are both `pull_request`-triggered and self-hosted, and both are fixed.
`release.yml`, `unity-benchmarks.yml` and `runner-bootstrap.yml` are self-hosted but never run on
`pull_request`, so no head guard applies to them. `deploy-docs.yml` keeps
`cancel-in-progress: false` on a GitHub-hosted runner with no head guard and no scarce resource.

## Honest residual

Two shapes of superseded leg remain, and neither is new:

- A leg **queued** when the head moves still takes a runner slot for about eleven seconds before
  `Require current PR head before setup` aborts it. That is the issue's third acceptance
  criterion met in substance rather than literally.
- A leg **already running** when the head moves finishes normally. On this pull request the
  superseded perf leg had acquired the organization lock two minutes before the push and ran its
  full benchmark. That is not a gap -- it is the constraint this issue puts on any fix.

There is also a new accumulation to be honest about. `cancel-in-progress: false` used to collapse
pending runs, so a group held at most one in-flight run plus one pending one no matter how many
pushes landed. Per-head groups remove that cap: N pushes now materialize N runs, and the doomed
leg count grows with the push count instead of staying at four. Two pushes, the common case,
dispatch exactly as many legs as before; the three pushes on this pull request produced eight
doomed Unity legs instead of four, about 96 seconds of runner time instead of 48, against the
10-33 minutes of `pending` removed.

Having the doomed run cancel itself would cap it again -- one doomed dispatch per superseded run
rather than four -- but it needs `actions: write` on the job that also carries the Unity serial
and the build-lock App key, and it puts a cancellation path into the workflow whose whole design
is that no cancellation path exists. Eight minutes of doomed-leg runner time across 120 runs does
not buy that. Filed as #406 with the measured numbers and explicit revisit conditions, so the
trade is not rediscovered by accident once this issue closes.

## Docs

`docs/runbooks/required-checks.md` still said a superseded run "keeps the concurrency group
through every matrix leg and the current head cannot start", which is the behavior this removes.
`.llm/skills/github-workflow-consistency/SKILL.md` records the rule, since its blanket
`cancel-in-progress: true` line did not describe the licensed Unity workflows.

## Verification

- `npm test` 421 pass; `validate:all`, `format:check`, `lint:markdown`, `check:spelling`,
  `actionlint`, `yamllint` and the pre-commit hook set all pass.
- `validate:unity-pr-policy` gained a per-head concurrency invariant covering **every**
  concurrency block in both workflows. Proved by mutation: reverting either group in either file
  to a ref-only key is rejected, a concurrency block with no group at all is rejected, and a
  `group:` key outside a concurrency block is correctly ignored. The workflow-level revert is
  additionally wired into the existing `require_policy_mutation_rejected` harness.
- Live supersession experiment on this pull request: see the comment below.

Closes #332

### DIFF
--- a/.github/workflows/perf-numbers.yml
+++ b/.github/workflows/perf-numbers.yml
@@ -69,8 +69,14 @@ on:
 # this workflow do not race each other. A separate freshness check still guards
 # against unrelated default-branch pushes landing while the benchmarks run. We do
 # NOT cancel in-progress runs -- that would abort an in-flight benchmark.
+#
+# Pull-request runs add the head SHA, so that merge-path serialization does not
+# also queue a pull request's current head behind a superseded run whose only
+# remaining act is to abort on the per-leg head guard (#332). Benchmarks stay
+# mutually exclusive through the organization build lock, and every licensed leg
+# re-checks the pull-request head before it can reach that lock.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number && format('{0}-{1}', github.event.pull_request.number, github.event.pull_request.head.sha) || github.ref }}
   cancel-in-progress: false
 
 # Top-level least-privilege. The default-branch refresh job (commit-perf-doc)
@@ -152,7 +158,12 @@ jobs:
       actions: read
       contents: read
     concurrency:
-      group: ${{ github.workflow }}-runner-preflight-${{ github.ref }}
+      # Keyed exactly like the workflow-level group. A `github.ref`-only key was
+      # unreachable while that group serialized every run of a pull request; now
+      # that it does not, the key has to match, or a newer run would cancel a
+      # superseded run's preflight, skip that run's benchmarks, and leave its
+      # aggregate reporting a result shape the gate below does not describe.
+      group: ${{ github.workflow }}-runner-preflight-${{ github.event.pull_request.number && format('{0}-{1}', github.event.pull_request.number, github.event.pull_request.head.sha) || github.ref }}
       cancel-in-progress: true
     outputs:
       has-autocommit-app: ${{ steps.check-autocommit-app.outputs.has-app }}
--- a/.github/workflows/unity-tests.yml
+++ b/.github/workflows/unity-tests.yml
@@ -13,7 +13,18 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: unity-tests-${{ github.event.pull_request.number || github.ref }}
+  # Push runs share one group per ref, so consecutive default-branch pushes stay
+  # serialized. Pull-request runs add the head SHA. Serializing those bought
+  # nothing and cost the head that matters: `cancel-in-progress: false` below is
+  # deliberate (a hard cancel mid-lease is the seat leak the four-layer
+  # license-return guarantee exists to prevent), so a superseded run held the
+  # shared group while its remaining legs waited on a scarce self-hosted runner
+  # purely to abort on the per-leg head guard, and the run for the current head
+  # sat pending with zero jobs for as long as that took -- 33 minutes on #402.
+  # Licensed Unity work is serialized by the organization build lock, not by this
+  # group, and every leg re-checks the pull-request head before it can reach that
+  # lock, so two runs of one pull request can never both do licensed work (#332).
+  group: unity-tests-${{ github.event.pull_request.number && format('{0}-{1}', github.event.pull_request.number, github.event.pull_request.head.sha) || github.ref }}
   cancel-in-progress: false
 
 permissions:
@@ -26,16 +37,15 @@ jobs:
   head-check:
     name: Unity head freshness
     # A second push to a pull request queues a second full matrix and does not
-    # retire the first. `cancel-in-progress: false` above is deliberate -- a
-    # hard cancel mid-lock is the exact scenario the four-layer license-return
-    # guarantee exists to avoid -- so the superseded run keeps the concurrency
-    # group until it has burned through every leg, and the run for the commit
-    # that matters cannot start. `Unity CI Success` is then absent on the
-    # current head, which reads as a blocked pull request on a check nobody
-    # misconfigured.
+    # retire the first: `cancel-in-progress: false` above is deliberate, because
+    # a hard cancel mid-lock is the exact scenario the four-layer license-return
+    # guarantee exists to avoid. The head-SHA-keyed concurrency group above lets
+    # the new run start anyway, so what is left to avoid is the superseded run
+    # handing a scarce self-hosted runner to every remaining leg purely so that
+    # leg can abort.
     #
     # Deciding it here, on GitHub-hosted infrastructure and before the lock is
-    # in reach, costs a superseded run one cheap job instead of nine
+    # in reach, costs a superseded run one cheap job instead of four
     # self-hosted ones. The per-leg `Require current PR head before setup`
     # guard stays: it covers the push that lands after this job has already
     # passed.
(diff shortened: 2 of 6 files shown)


wallstop:
## Live supersession test on this pull request

Pushed `93e1c1f9` at **03:15:58Z** while `8c159f54` was still in flight in exactly the #332
shape: `head-check` and `runner-preflight` both green, all four `Unity Tests` legs `queued` with
no runner assigned, and one `Performance Numbers` leg already `in_progress` since 03:13:31Z.

| | Superseded run (`8c159f54`) | Current head (`93e1c1f9`) |
| --- | --- | --- |
| `Unity Tests` run | `31765995889`, created 03:10:35Z | `31766286569`, created **03:16:05Z** |
| First job created | 03:10:35Z | **03:16:05Z** (+7s from push) |
| `Unity head freshness` | success | success at **03:16:07Z** (+9s) |
| `runner-preflight` | success | success |
| Matrix legs created | 03:10:52Z | **03:16:22Z** (+24s) |
| `Performance Numbers` run | `31765995867`, 1 leg `in_progress` | `31766286537`, 4 jobs created |

The current head had a live, reporting `Unity Tests` run **24 seconds** after the push, with all
four legs scheduled, while the superseded run was still sitting on its own four queued legs.

Three things to read off that:

1. **The `pending`-with-zero-jobs state is gone.** This is the exact condition recorded on #332
   for PR #402, where the superseded run held the group and the current head's `Unity Tests` and
   `Performance Numbers` both sat `pending` with no jobs created, so `gh pr checks` showed only
   the static checks. Both workflows now materialize immediately.
2. **Nothing was cancelled.** The superseded run's in-flight `Perf benchmarks` leg, running since
   03:13:31Z, was left alone to finish its lease. That is the constraint #332 puts on any fix, and
   here it holds by construction rather than by a check, because no cancellation path exists.
3. **The two runs cannot both do licensed work.** The superseded run's Unity legs abort on
   `Require current PR head before setup`, and its perf leg is re-checked before lock acquisition.

The residual is visible too: the superseded run's four legs each still take a runner slot for
about eleven seconds before aborting. That is the honest gap against the issue's third acceptance
criterion, and it is 44 seconds against the 33 minutes this removes.

PR #415, opened 2026-08-14
Fix post-processor snapshot consistency

### LINKED ISSUE #413: bug: a post-processor registered from an interceptor runs this emission only when the post sink was empty
### Description

Whether a post-processor registered from inside an interceptor runs in the **same** emission depends
on something the caller cannot see: whether that message type already had at least one
post-processor registered when the emission started.

- **Sink already populated** -> the late post-processor waits for the next emission.
- **Sink empty** -> the late post-processor runs in the emission that registered it.

The same user action produces two different behaviours, and the second one contradicts the
documented contract in
[`docs/concepts/interceptors-and-ordering.md`](https://github.com/Ambiguous-Interactive/DxMessaging/blob/master/docs/concepts/interceptors-and-ordering.md),
which says a listener added during an emission is not invoked for the current message and that this
"applies to all registration types: handlers, interceptors, and post-processors."

### Steps to Reproduce

Both cases register a post-processor from inside an interceptor. Only the second one has a
pre-existing post-processor.

```csharp
// Case A - empty post sink. The late post-processor RUNS this emission.
_ = token.RegisterUntargetedInterceptor<Msg>((ref Msg _) =>
{
    if (!added) { added = true; _ = token.RegisterUntargetedPostProcessor<Msg>((ref Msg __) => late++); }
    return true;
});
_ = token.RegisterUntargeted<Msg>((ref Msg _) => { });
bus.UntargetedBroadcast(ref message);
// late == 1

// Case B - one post-processor already registered. The late one WAITS.
_ = token.RegisterUntargetedPostProcessor<Msg>((ref Msg _) => existing++);
_ = token.RegisterUntargeted<Msg>((ref Msg _) => { });
_ = token.RegisterUntargetedInterceptor<Msg>((ref Msg _) =>
{
    if (!added) { added = true; _ = token.RegisterUntargetedPostProcessor<Msg>((ref Msg __) => late++); }
    return true;
});
bus.UntargetedBroadcast(ref message);
// late == 0
```

Case B is what `UntargetedPrefreezeTests.PostProcessorRegisteredByInterceptorDoesNotFireSameEmission`
pins. Case A is untested.

### Expected Behavior

One rule regardless of prior sink population. Following the documented snapshot contract, a
post-processor registered during an emission should first run on the **next** emission in both
cases.

### Actual Behavior

Case A runs it immediately; case B defers it.

### Additional Context

This is longstanding, not a regression: verified on `master` at `d2f5d65b` and on
#412, which behaves identically here (#412 pins both variants so the difference cannot drift
silently).

The mechanism is the pre-freeze in `MessageBus.UntargetedBroadcast`. The post snapshot is acquired at
emission start only when `plan.scalarPost` is non-null with a non-empty sink. When it is empty, no
snapshot exists, and the post phase's live re-check then acquires a fresh snapshot that includes the
mid-emission registration. The behaviour is deliberate at the code level -- the method comment says a
post-processor registered "into a previously-snapshotless sink mid-emission is observed here" -- but
it was never reconciled with the documentation, and the two branches are not distinguishable from a
caller's perspective.

Fixing it means picking one:

1. **Always defer** (matches the docs). The live re-check would need to ignore sinks that had no
   snapshot at emission start, which likely means recording "there were no post-processors at
   emission start" rather than relying on `IsInitialized` as the proxy.
2. **Always observe.** Simpler in code but contradicts the documented contract and the frozen-listener
   guarantee users rely on for loop safety.

Option 1 looks right, but it changes observable behaviour for anyone relying on case A today, so it
wants a deliberate decision rather than being folded into a perf change.

The targeted and broadcast lanes have the same shape (`RunTargetedPostPhases`, and the source-keyed
re-resolution in `SourcedBroadcast`) and should be swept together.

wallstop:
Closes #413

## Summary

- Freeze post-processor membership at emission start even when a route was initially empty.
- Preserve the emission-start post snapshot when an interceptor rewrites a target or source, including processors removed during the emission.
- Keep transient rewritten-route history bounded across nested emissions, release superseded pooled snapshots, reuse the compaction index, and reclaim idle holders through force trim.
- Add targeted, broadcast, untargeted, reset, nesting, exception, allocation, and reclamation coverage.
- Correct the test-suite performance runbook so it does not overstate the runtime-only wall-clock guard.

## Root cause

Initially empty post sinks were looked up live after interceptors and handlers ran, while populated sinks used a prefrozen snapshot. Rewritten target/source routes were also resolved from live caches. That made late registrations visible only for some route shapes and made live-cache reconstruction unable to retain a processor removed during the current emission.

## Impact

All post-processor surfaces now follow one rule: additions made during an emission wait until the next emission, while removals still finish the current frozen post phase.

## Validation

- Three-way adversarial semantics, performance/memory, and tests/docs review: clean consensus
- Unity 6000.4.6f1 focused retargeting fixture: 34/34 twice
- Full PlayMode runtime: 1091/1091 twice, with 1 expected skip
- Rewritten target/source allocation rows: 2/2
- Full EditMode editor + allocation assemblies: 1024 passed, 1 skipped, 8 established inconclusive probes
- Node tests: 421/421
- Source-generator tests: 246/246
- Documentation tests: 583/583
- `npm run validate:all`, analyzer reproducibility, CSharpier, Prettier, markdownlint, cspell, and diff checks: clean

---

> [!NOTE]
> **High Risk**
> Changes core `MessageBus` dispatch semantics for all post-processor surfaces (empty routes, fast path, interceptor retargeting, reset), with behavior shifts users may rely on; mitigated by extensive regression tests.
> 
> **Overview**
> Fixes **#413** by making **post-processor dispatch** follow one frozen-snapshot rule everywhere: membership is fixed at **emission start**, so **additions during an emission wait until the next one**, while **removals still run for the current post phase**.
> 
> **`MessageBus`** stops **live post-sink lookups** after handlers/interceptors (including the **fast-path** fallback when the dispatch plan changes mid-emission). **Untargeted** and **featured** paths only dispatch **prefrozen** post snapshots, gated by **reset generation** so `Reset()` mid-dispatch does not run stale post work.
> 
> When an **interceptor rewrites** target or source, rewritten **context post routes** resolve through **pre-mutation snapshot history** (`CaptureContextPostRouteBeforeMutation` / `TryGetContextPostRouteAtEmissionStart`) instead of live caches—so a processor **removed during the emission** still runs, and one **added** does not. History is **compacted on nested dispatch exit**, **released on outermost lease teardown**, and **trimmed** when oversized.
> 
> **CHANGELOG**, **interceptors-and-ordering** docs, a **test-suite performance runbook** correction, and **tests** (retargeting, mutations, reset, allocations, nesting/memory) lock in the new behavior—including reversing the old “last post-processor deregistered skips same emission” edge case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b3b4c824c5ee91df7cfe9c44c8feace1f945c98. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Runtime/Core/MessageBus/MessageBus.cs
+++ b/Runtime/Core/MessageBus/MessageBus.cs
@@ -399,6 +399,114 @@ namespace DxMessaging.Core.MessageBus
         }
 
         /// <summary>
+        /// Captures one context-keyed post route immediately before a registration mutation.
+        /// The first mutation after an emission began owns the exact route snapshot that emission
+        /// must observe if an interceptor later rewrites the target/source to this context.
+        /// Entries live only until the outermost dispatch lease exits.
+        /// </summary>
+        private readonly struct PostRouteSnapshotHistoryEntry
+        {
+            public PostRouteSnapshotHistoryEntry(
+                long resetGeneration,
+                int messageTypeIndex,
+                SlotKey slotKey,
+                InstanceId context,
+                long captureEmissionId,
+                int captureDispatchDepth,
+                long mutationTick,
+                DispatchSnapshot snapshot
+            )
+            {
+                this.resetGeneration = resetGeneration;
+                this.messageTypeIndex = messageTypeIndex;
+                this.slotKey = slotKey;
+                this.context = context;
+                this.captureEmissionId = captureEmissionId;
+                this.captureDispatchDepth = captureDispatchDepth;
+                this.mutationTick = mutationTick;
+                this.snapshot = snapshot;
+            }
+
+            public readonly long resetGeneration;
+            public readonly int messageTypeIndex;
+            public readonly SlotKey slotKey;
+            public readonly InstanceId context;
+            public readonly long captureEmissionId;
+            public readonly int captureDispatchDepth;
+            public readonly long mutationTick;
+            public readonly DispatchSnapshot snapshot;
+
+            public PostRouteSnapshotHistoryEntry ReassignCapture(long emissionId, int dispatchDepth)
+            {
+                return new PostRouteSnapshotHistoryEntry(
+                    resetGeneration,
+                    messageTypeIndex,
+                    slotKey,
+                    context,
+                    emissionId,
+                    dispatchDepth,
+                    mutationTick,
+                    snapshot
+                );
+            }
+        }
+
+        private readonly struct PostRouteKey : IEquatable<PostRouteKey>
+        {
+            private readonly int _messageTypeIndex;
+            private readonly SlotKey _slotKey;
+            private readonly InstanceId _context;
+            private readonly int _hash;
+
+            public PostRouteKey(int messageTypeIndex, SlotKey slotKey, InstanceId context)
+            {
+                _messageTypeIndex = messageTypeIndex;
+                _slotKey = slotKey;
+                _context = context;
+                unchecked
+                {
+                    int hash = 17;
+                    hash = (hash * 31) + messageTypeIndex;
+                    hash = (hash * 31) + slotKey.GetHashCode();
+                    hash = (hash * 31) + context.GetHashCode();
+                    _hash = hash;
+                }
+            }
+
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public bool Equals(PostRouteKey other)
+            {
+                return _hash == other._hash
+                    && _messageTypeIndex == other._messageTypeIndex
+                    && _slotKey == other._slotKey
+                    && _context == other._context;
+            }
+
+            public override bool Equals(object obj)
+            {
+                return obj is PostRouteKey other && Equals(other);
+            }
+
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public override int GetHashCode()
+            {
+                return _hash;
+            }
+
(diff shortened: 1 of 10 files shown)


Runtime/Core/MessageBus/MessageBus.cs
```diff
@@ -4482,28 +4581,32 @@ out HandlerCache<int, HandlerCache> sortedHandlers
...
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private bool RunTargetedPostPhases<TMessage>(
             ref InstanceId target,
             InstanceId preInterceptorTarget,
             ref TMessage typedMessage,
-            long planVersion,
             DispatchSnapshot targetedPostSnapshot,
             DispatchSnapshot targetedWithoutTargetingPostSnapshot,
-            long emissionId
+            long emissionId,
```
wallstop, line 4597: Is it possible/more performant to create like, a struct with all of these parameters and just pass it by ref, to share between all the input stuff? Or would that be less efficient? If out-of-scope, file as a follow-up research item

PR #456, opened 2026-08-23
Fire base-call hide diagnostics only for real hides and report each once

### LINKED ISSUE #453: bug(analyzer): DXMSG009 fires for hiding OnApplicationPause, which MessageAwareComponent does not declare
## What happens

A component deriving from `MessageAwareComponent` that declares its own `OnApplicationPause` gets:

> DXMSG009: `X` declares OnApplicationPause without 'override' or 'new'; this implicitly hides MessageAwareComponent.OnApplicationPause (CS0114) and the messaging system will not function.

`MessageAwareComponent` does not declare `OnApplicationPause`. `Runtime/Unity/MessageAwareComponent.cs` declares `OnEnable`, `OnDisable`, `OnDestroy` and `OnApplicationQuit` — and nothing else in that family.

## Where it comes from

`Editor/Analyzers/BaseCallTypeScannerCore.cs:59`, `:73`, `:121` put `OnApplicationPause` (and `OnApplicationFocus`) into the guarded-method set. The comment at `:45` says they are guarded **prospectively** — i.e. against a base method that might exist one day.

So the diagnostic asserts something untrue about the current shipped base class, and its remedy text ("add `override`") names a method that cannot be overridden. There is no `CS0114` here because there is nothing to hide.

## It is also emitted twice

For one declaration, the analyzer reports the diagnostic twice at the same location.

## Why it cannot just be lived with

Adding `new` to satisfy the message trades DXMSG009 for `CS0109` ("does not hide an inherited member"), which is the compiler correctly saying the diagnostic was wrong. So there is no spelling of the method that is clean, and the warning becomes a permanent entry in any consumer's warning baseline. Ours has carried it for months, at count 2, with a paragraph of documentation explaining that it cannot be lowered.

## Ask

- Gate the guarded-method set on the methods `MessageAwareComponent` **actually declares**, rather than on a prospective list. If the intent is to reserve the two hooks, the cleaner route is to add them to `MessageAwareComponent` so the diagnostic becomes true.
- De-duplicate the report so one declaration produces one diagnostic.

A false positive that has no clean suppression is worse than a missing check, because it trains every consumer to carry a baseline entry — and a baseline entry is where a real one hides later.

Found while sweeping our tree for local workarounds around first-party packages (Ambiguous-Interactive/IshoBoy#666).

wallstop:
## Why

`MessageAwareComponent` declares neither `OnApplicationFocus(bool)` nor `OnApplicationPause(bool)`, but both names sat in the guarded set. A subclass declaring its own `OnApplicationPause(bool)` got DXMSG009 claiming the method "implicitly hides MessageAwareComponent.OnApplicationPause (CS0114)" - a hiding that does not exist. No spelling of the method was clean: adding `new` traded DXMSG009 for compiler warning CS0109, so consumers carried a permanent warning-baseline entry.

The same declaration was also reported twice. When this analyzer DLL loads twice into one compilation - typically a stale in-project copy beside the package's RoslynAnalyzer-labeled payload - each registration reports its own copy of every diagnostic at the same span.

Found in Ambiguous-Interactive/IshoBoy#666; fixes #453.

## What

- Add `HidesInheritedMember`: a walk of the base chain looking for an ordinary method with the same name, generic arity, and parameter list (return type excluded, matching C# hiding rules). DXMSG009 and DXMSG007 now fire only when it returns true, so both stay silent for guarded names no ancestor declares and resume automatically if the base class adds these hooks. The prospective entries stay in `GuardedMethodNames`, the consequence-text dictionaries, and the IL scanner sets; the scanner already gated on real base virtuals and needed only comment updates.
- Route all seven diagnostic report sites through `ReportOnce`, keyed per compilation by diagnostic id plus source span. The compiler host loads an assembly once per identity, so duplicate registrations share the table and collapse to one report per location. State is scoped to the compilation, so nothing leaks across compiles; RS1008 is suppressed at the field because the weak table is the leak-safe form of the pattern the rule wants.
- Document the new trigger conditions in `docs/reference/analyzers.md` (DXMSG007/DXMSG009 sections, signature filter, a short "One diagnostic per declaration" note) and update the base-call-contract skill wording.

## How I know it is correct

- New analyzer tests pin each behavior: bool and zero-arg declarations of both hook names stay silent; `new` on undeclared names stays silent with CS0109 allowed; DXMSG009 and DXMSG006 resume once an intermediate base declares the hook; double registration yields exactly one diagnostic.
- Full dotnet suite: 255 passed, 0 failed. Docs snippet/reference tests: 601 passed.
- Local Unity leg over unity-mcp (6000.4.6f1): refresh imported the refreshed payload with zero console errors; EditMode run twice: 805 pass / 0 fail / 1 skip / 8 inconclusive both times. The inconclusives are `SampleQualityContractTests` fixture-gated contracts that are inconclusive by design without the CI-imported samples folder.
- `npm run check:analyzers` reports the shipped payload reproducible from source.

---

> [!NOTE]
> **Medium Risk**
> Changes compile-time analyzer reporting for lifecycle methods, which can hide or surface warnings in consumer projects. Logic is well-tested and does not change runtime messaging behavior.
> 
> **Overview**
> Stops **DXMSG007** and **DXMSG009** from claiming a CS0114 hide when a subclass declares `OnApplicationFocus(bool)` or `OnApplicationPause(bool)`. `MessageAwareComponent` does not declare those hooks, so the old warning was false and adding `new` only swapped it for CS0109.
> 
> Hide diagnostics now require `HidesInheritedMember`: an ancestor must actually declare a matching virtual/abstract method. Prospective names stay in the guarded set so coverage resumes automatically if the base class adds the hooks. `static new` is also silent because it cannot join the override chain.
> 
> Reports each diagnostic once per (id, source span) per compilation via `ReportOnce`, so a duplicate analyzer registration (stale in-project DLL plus package payload) no longer doubles every warning.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 505b23e7f65eb3b48c738749c7aa7c6a1bf30e74. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/reference/analyzers.md
```diff
@@ -193,7 +193,7 @@ See [Suppression precedence](#suppression-precedence) for the full ordering.
 
 - **Severity:** Warning
 - **Source:** `MessageAwareComponentBaseCallAnalyzer`
-- **Triggered when:** A subclass uses the `new` modifier (instead of `override`) on one of the five guarded method names.
```
wallstop, line 196: Can you scan our public documentation (github pages) + comments to ensure that there is no where else that references these special methods?

wallstop: Swept every tracked surface for `OnApplicationFocus` / `OnApplicationPause` / "prospective" and checked each remaining mention against the new gate. Result: no other public documentation or user-visible text claims these hooks are declared, guarded as overrides, or produce diagnostics.

Public documentation (the GitHub Pages source tree):

- `docs/reference/analyzers.md` is the only page naming them. Both mentions (DXMSG009 trigger conditions and the signature-filter section) now state the real contract: recognized names, guarded prospectively, silent until an ancestor declares a virtual member with a matching signature, resuming automatically if the base class adds the hooks.
- The other pages that say "five guarded methods" (quick-start, getting-started, troubleshooting, faq) describe DXMSG006 override coverage of the five methods `MessageAwareComponent` actually declares - accurate, untouched.
- `README.md`, `llms.txt`, `docs/overrides/*.html`, and `Samples~/` contain no references to either hook.

Code comments:

- The analyzer's `GuardedMethodsWithBoolSignature` doc, the DXMSG009 gate comment, and the scanner-core header already stated the gating; this commit also aligned the two consequence-dictionary comments that still implied firing was possible today (`MessageAwareComponentBaseCallAnalyzer.cs`, `BaseCallTypeScannerCore.cs`).
- The DXMSG009 descriptor description and message format never listed the two hooks and remain accurate unchanged.

Generated agentic-knowledge surfaces (`.claude/skills/`, `.agents/skills/`) are thin pointers regenerated by `npm run llm:index`; the source skill wording says a virtual member is required, and `npm run llm:check` passes.

Comment-only edits do not change the compiled payload: rebuilding reproduced byte-identical DLLs in `Runtime/Analyzers/`. Suite stays 257/257.

PR #496, opened 2026-08-30
Fix reflexive routing and destroyed listener cleanup

### LINKED ISSUE #489: bug: combined ReflexiveSendMode flags apply the wrong active-state filter
## Problem

`ReflexiveSendMode.OnlyIncludeActive` is documented as an independent flag, but combined traversal modes do not honor that contract.

- `Upwards | Downwards`, `Upwards | Flat`, and other fallback combinations call the manual dispatcher with `onlyActive: true` even when `OnlyIncludeActive` is absent. Disabled receivers are silently skipped.
- `Flat | OnlyIncludeActive` checks only `MonoBehaviour.enabled`. An enabled component on an inactive GameObject still receives the call even though `isActiveAndEnabled` is false.
- The downward combined-mode fallback uses the overload of `GetComponentsInChildren` that excludes inactive descendants regardless of the flag.

## Root cause

`MessageBus.TargetedBroadcast` selects Unity's native fast path only for one direction without `OnlyIncludeActive`. Every combined-mode fallback then hardcodes active-only delivery instead of deriving it from the flag, and the helper uses `enabled` instead of `isActiveAndEnabled`.

Relevant code: `Runtime/Core/MessageBus/MessageBus.cs`, reflexive dispatch around the Upwards, Downwards, and Flat branches plus the private `SendMessage` helper.

## Impact

Hierarchy receivers can miss gameplay notifications when callers combine direction flags, and inactive GameObjects can receive notifications explicitly requested as active-only.

## Acceptance criteria

- Combined direction modes deliver to disabled and inactive receivers when `OnlyIncludeActive` is absent.
- Every traversal mode excludes disabled components and components on inactive GameObjects when `OnlyIncludeActive` is present.
- Downward traversal includes inactive descendants only when active filtering is absent.
- Duplicate suppression across combined directions remains intact.
- Data-driven Unity tests cover combined directions, disabled components, and inactive GameObjects.

### LINKED ISSUE #490: bug: missing zero- and one-argument reflexive receivers log Unity errors
## Problem

A reflexive call to a method that does not exist is silent for two or more arguments, but zero- and one-argument calls log Unity `RequireReceiver` errors.

That makes behavior depend on argument count and turns optional reflected delivery into an unexpected console error.

## Root cause

The zero- and one-argument fast paths call `GameObject.SendMessage`, `SendMessageUpwards`, and `BroadcastMessage` overloads without `SendMessageOptions.DontRequireReceiver`. Unity therefore uses `RequireReceiver`. The custom two-or-more-argument path caches a missing method as a null dispatcher and silently no-ops.

Relevant code: `Runtime/Core/MessageBus/MessageBus.cs` reflexive Upwards, Downwards, and Flat native fast paths.

## Impact

Optional gameplay notifications can fail Unity tests or pollute the player log solely because their payload has zero or one argument.

## Acceptance criteria

- Missing reflexive receivers are silent no-ops for zero, one, and multiple arguments.
- Flat, Upwards, and Downwards modes behave consistently.
- Existing matching receivers still receive zero- and one-argument messages.
- Data-driven Unity tests cover every direction and argument-count boundary without accepting unexpected logs.

### LINKED ISSUE #491: bug: MessagingComponent cannot release a destroyed manual listener
## Problem

`MessagingComponent.Release(listener)` returns `false` after a manually managed listener has been destroyed, even when that exact managed object remains the key in the component's listener dictionary.

The token is not disposed, the dictionary entry remains, and its bus registrations can survive.

## Root cause

`Release` begins with `if (listener == null)`. Unity overloads equality, so a destroyed `MonoBehaviour` compares equal to null even though the managed reference is non-null and is still the dictionary key. The method returns before `TryGetValue`, disposal, and removal.

`Create` should keep lifetime-aware Unity null validation. Cleanup needs reference-null validation so it can release retained destroyed keys.

Relevant code: `Runtime/Unity/MessagingComponent.cs:192-203`.

## Impact

Manual listener lifecycles can retain destroyed Unity objects, tokens, and active registrations. This is a memory and registration leak and can leave stale callbacks reachable.

## Acceptance criteria

- `Release` accepts a destroyed-but-managed listener reference when it matches a retained key.
- The token is disposed and the listener entry is removed.
- Bus registration counters return to their baseline.
- A true CLR null still returns `false`.
- Unknown live and unknown destroyed listeners still return `false`.
- Disposal failure keeps the entry retryable, preserving the existing contract.

wallstop:
Closes #489.
Closes #490.
Closes #491.

## Combined reflexive modes route the wrong receivers (#489)

**Why.** Combined modes always filtered disabled components. Downward sends also skipped inactive descendants without `OnlyIncludeActive`. Nested sends could overwrite the outer traversal state.

**What changed.** Combined traversal now follows each mode flag. Dispatch-scoped pooled state isolates nested sends. Reset and destroyed-object guards stop unsafe traversal.

**How we know.** Direction, activity, nesting, reset, destruction, pool reuse, trim, and settings tests cover these paths.

## Missing reflexive receivers log Unity errors (#490)

**Why.** Native zero-argument and one-argument sends required a receiver. Unity logged errors when no matching method existed.

**What changed.** Native sends now make receivers optional. Active single-direction sends keep Unity's argument compatibility. Other sends use exact manual matching.

**How we know.** Parameterized tests cover every direction with zero, one, and two arguments. Compatibility tests cover ignored and assignable arguments.

## Destroyed listeners cannot be released (#491)

**Why.** Unity's null operator hid destroyed listener references before dictionary lookup. Their tokens stayed registered and retained.

**What changed.** `Release` now rejects only a CLR null reference. A destroyed retained listener can reach normal token cleanup.

**How we know.** Lifecycle tests cover live, destroyed, unknown, failed, and retried releases. Leak checks cover all registration counters.

## Validation

The script suite passes 528 tests. Source-generator tests pass 257 tests. Documentation tests pass 616 tests. Repository validation passes.

Two final adversarial reviews report zero findings.

The shared Unity host contains stale external sample assets and cannot compile this branch. Generated clean CI hosts remain authoritative.

---

> [!NOTE]
> **Medium Risk**
> Touches `MessageBus` dispatch and reflexive traversal semantics on a hot path, but changes are regression-focused with extensive new tests and documented v4.0.0 behavior.
> 
> **Overview**
> Fixes **reflexive `SendMessage` routing** and **listener cleanup** on the bus, plus a small **LLM index** hygiene change.
> 
> **Reflexive sends (#489, #490).** Combined `ReflexiveSendMode` flags now traverse each enabled direction with correct active/inactive filtering (`OnlyIncludeActive` is opt-in; inactive descendants are included otherwise). Nested reflexive emissions use a scoped `ReflexiveDispatchLease` and pooled `ReflexiveDispatchState` so inner sends cannot clobber outer traversal or deduplication. Traversal skips destroyed receivers, honors reset generation, and routes 0/1-argument single-mode sends through Unity native `SendMessage` with `DontRequireReceiver` so missing methods stay silent. `DispatchLease` and reflexive leases track **bus-held lease IDs** so copied struct disposes cannot double-unwind dispatch depth.
> 
> **`MessagingComponent.Release` (#491).** Lookup uses `ReferenceEquals` instead of Unity’s overloaded null check, so a retained reference to a destroyed listener can still dispose its token and drop the cache entry.
> 
> **Tooling/docs.** Stops generating and committing `.llm/index.json` (gitignored; harness tests updated). Skills and user docs note copy-safe disposable leases, reflexive behavior, and the new per-bus reflexive pool in trim/settings.
> 
> **Tests.** Broad reflexive, nested/destroy/reset, lease-copy, lifecycle release, and memory-reclamation coverage back the behavior changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e9963b1095aad417e5cbe7efdc8904aa9d7cd782. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Core/MessageBus/MessageBus.cs
```diff
@@ -4465,158 +4564,79 @@ ref typedMessage
...
                     if (sendMode.HasFlagNoAlloc(ReflexiveSendMode.Downwards))
                     {
-                        if (
-                            !sendMode.HasFlagNoAlloc(ReflexiveSendMode.Upwards)
-                            && !sendMode.HasFlagNoAlloc(ReflexiveSendMode.Flat)
-                            && !sendMode.HasFlagNoAlloc(ReflexiveSendMode.OnlyIncludeActive)
-                        )
+                        List<Transform> transforms = _reflexiveDispatchState.transforms;
+                        transforms.Clear();
+                        go.GetComponentsInChildren(true, transforms);
```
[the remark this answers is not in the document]
wallstop: Fixed in `c3b16e39`. Downward enumeration now checks the original GameObject through Unity's fake-null operator after upward delivery. The parameterized regression covers both Upwards and combined Upwards|Downwards when the origin destroys itself.

Runtime/Core/MessageBus/MessageBus.cs
```diff
@@ -798,6 +798,65 @@ public void Dispose()
...
+                bus._reflexiveDispatchDepth++;
+                bus._reflexiveDispatchState.Clear();
+            }
+
+            public void Dispose()
+            {
+                MessageBus bus = _bus;
+                ReflexiveDispatchState completedState = bus._reflexiveDispatchState;
+                completedState.Clear();
+                bus._reflexiveDispatchDepth--;
```
wallstop, line 878: Seems like we copied leases can decrement dispatch depth on double dispose?

wallstop: Good catch. Fixed in `17362677`. Both `DispatchLease` and `ReflexiveDispatchLease` now use owner-held opaque identities, so copied, stale, or out-of-order values cannot unwind an active scope twice. The regression matrix covers both disposal orders, stale reuse, and nested out-of-order disposal for both lease types. I also swept the other value-type disposables: the public global-bus scope and registration wrapper already use copy-safe identities, and the allocation probe is naturally idempotent. The collection-pooling and test-coverage skills now require this copy-safety contract.

.llm/index.json
```diff
@@ -386,7 +386,7 @@
         "category": "testing",
         "tags": "testing, coverage, edge-cases, data-driven, unity, nunit, best-practices, quality"
       },
-      "lineCount": 77,
+      "lineCount": 82,
```
wallstop: We should not include data that can be derived from files (single source of truth)

wallstop: Agreed. Fixed in `e9963b10`. The repository had no consumer for `.llm/index.json`, and every field was derived from canonical `SKILL.md` frontmatter or directory contents, so I removed the entire JSON artifact and its generator API rather than only dropping `lineCount`. The human skill index and client discovery mirrors remain because they are consumed. I also added an ignore rule, updated the skill-maintenance guidance to prohibit derived machine manifests, and updated the harness tests. Full `npm test` (528/528) and `npm run validate:all` pass.

Issue #509, opened 2026-08-31
test: build a differential bus state-machine oracle


wallstop:
Part of #497. Uses the evidence format from #508. Blocks runtime candidates in #501 through #505.

## Question

Can a candidate bus implementation be shown trace-equivalent to the control over generated sequences of dispatch, mutation, reentrancy, reset, failure, and lifecycle operations?

## Mechanism

Example tests cover known cases but miss interactions and operation ordering. A differential state-machine runner can execute one deterministic sequence against control and candidate implementations, compare externally observable state, and shrink a failure without copying production routing logic.

## Contract

Observe production implementations; do not reimplement their algorithms. Compare callback order, message/context values, found result, diagnostics, exception outcome, registration counters, occupied slots, trim result, retained references, and subsequent usability.

Generate register, duplicate register, enable, disable, deregister, emit, nested emit, self-mutation, reset, trim, scene transition, host destruction, and throw operations. Preserve live activity, emission-start snapshots, deferred reset, exception cleanup, stale-handle safety, multiple-bus isolation, and all message kinds.

## Factors and workloads

- `MessageScenarios.AllKinds` through shared parameterization.
- Sequence length, seed, type count, fan-out, priorities, active density, and route cardinality.
- Same-kind and all six cross-kind reentrancy combinations through depth 10.
- Copied, duplicate, foreign-token, out-of-order, reused, stale, and generation-wrap handles.
- Scene unload, `DontDestroyOnLoad`, prefab churn, disable/re-enable, quit, and host destruction.
- Empty, boundary, invalid, unexpected, and throwing operations.
- Stable, same-type churn, and unrelated-type churn.

## Primary response

The number of observable trace mismatches between control and candidate.

## Independent unit

One deterministic seed and its complete operation sequence. Operations inside the sequence are correlated.

## Effect threshold

Behavioral acceptance is exact: zero mismatches across all preregistered seeds and canonical fixtures. One reproducible mismatch rejects a candidate regardless of throughput.

## RED proof

Supply test-only faulty adapters that independently drop a callback, reorder equal-priority callbacks, continue after reset, reuse a stale handle, skip exception cleanup, misreport diagnostics, retain a dead reference, and leak a registration. Detect and shrink each mutant.

## GREEN suites

[...]

Issue #508, opened 2026-08-31
perf lab: retain and replay content-addressed experiment evidence


wallstop:
Part of #500. Integrates with the frozen profile from #506.

## Question

Can a reviewer reproduce every reported effect, interval, and decision after ordinary CI artifacts expire, and detect any missing or changed input?

## Mechanism

The checked-in CSV is replaced and current large CI artifacts expire. Screenshots and copied summaries cannot establish provenance or reproduce a decision.

## Contract

Evidence is append-only and content-addressed. Every derived value traces to raw inputs, source, tree, player, profile, workload, environment, and analysis hashes.

- Commit compact manifests, results, decisions, and retrieval instructions under `progress/perf-experiments/<experiment-id>/`.
- Publish each sealed large bundle as an immutable GitHub prerelease named `perf-evidence-<experiment-id>-<manifest-sha8>`.
- Never replace an asset. A correction creates a new revision with a new manifest hash.
- Never retain credentials, Unity license material, access tokens, or unreviewed personal/device identifiers.

## Factors and workloads

Validate bundles for canonical paired throughput, allocation-heavy SubUnsub, cold latency, frame/queue latency, WPR/PMU/native mapping, and ARM64 energy captures. Exercise missing, duplicated, corrupted, reordered, schema-upgraded, inaccessible, and cross-platform-path cases.

## Primary response

A clean restore reproduces the normalized machine-readable effect, interval, validity checks, and decision from the retained bundle.

## Independent unit

One complete experiment bundle.

## Effect threshold

Acceptance is exact: every declared artifact verifies against SHA-256, every required input is present, and deterministic reducers reproduce normalized output exactly. One changed byte or undeclared input must fail verification.

## RED proof

- Modify one raw cycle, manifest, binary hash, or analysis file and prove replay fails.
- Remove a required artifact and prove replay names it.
- Make a GitHub prerelease asset inaccessible and prove the bundle is incomplete.
- Attempt to overwrite an experiment ID with different content and prove validation fails.
- Seed a credential fixture and prove publication refuses it.

## GREEN suites

Pass schema, hash, replay, path-portability, secret-scanning, reducer, renderer, and workflow fixture tests on supported CI operating systems. Independently restore at least one real bundle from every artifact class.

## Stop rule

Do not cite a result as campaign evidence if its bundle cannot replay. Mark a conclusion evidence-blocked when required large artifacts cannot be retrieved.

## Immutable evidence

[...]

Issue #506, opened 2026-08-31
perf lab: freeze the canonical IL2CPP profile and add a shipping-fidelity player


wallstop:
Part of #500.

## Question

Can every published result prove its effective IL2CPP configuration, and does a minimal stripped consumer preserve supported behavior without relying on test assemblies or accidental AOT roots?

## Mechanism

Build settings, stripping roots, symbols, native toolchains, and cached products change generic sharing, code layout, first-touch work, binary size, and throughput. The current test player is needed for discovery, but it is not a shipping-size or stripping oracle. The canonical profile now pins `OptimizeSpeed`; generated configuration logs and configured, prebuild, and postbuild evidence record the effective mode. Remaining provenance and campaign factors are tracked below.

## Contract

Maintain two separate profiles:

- Canonical verdict player: pinned Windows x64 Standalone IL2CPP Release, .NET Standard 2.1, with the test infrastructure required by the benchmark.
- Shipping-fidelity player: a minimal consumer with no test assemblies, production-compatible stripping, and explicit AOT roots.

Never substitute one profile's result for the other. Preserve Unity 2021.3 support, the public API, all message kinds, private and nested message support, untyped dispatch, zero-allocation warm dispatch, and current ordering, mutation, reset, reentrancy, exception, diagnostics, and reclamation behavior.

## Factors and workloads

- Unity 2021.3 and the current pinned Unity line.
- Clean and incremental builds as separate factors.
- Test player versus shipping-fidelity player.
- Every supported non-Disabled stripping level, including Minimal, Low, Medium, and High where present.
- Class and readonly-struct messages; public, private, and nested types.
- Typed and untyped dispatch.
- 1, 16, 256, and 1,000 closed message shapes.
- Cold startup, first registration, first dispatch, warm dispatch, trim, player size, and build time.

Record architecture, backend, API compatibility, IL2CPP compiler configuration and code-generation mode, stripping, exceptions, stack traces, Development/debug/profiler/deep-profile flags, assertions, incremental GC, test-assembly inclusion, editor/compiler/linker versions, native flags, symbols, and cache state.

## Primary response

Exact conformance against a versioned machine-readable profile identified by a SHA-256 hash.

## Independent unit

One clean IL2CPP build and its resulting player. Repeated launches of one player are nested observations.

## Effect threshold

[...]

PR #555, opened 2026-09-07
Expand lifecycle replay and retain native performance evidence

wallstop:
## Why

Generated replay lacks global override lease actions. Allocation observations and native build inputs also need evidence that can be checked later.

## What changed

- Add version 10 lease replay, copied-handle checks, generation boundaries, and cleanup coverage for #509.
- Preserve raw SubUnsub allocation output and require exact CSV replay for #508.
- Retain log-selected Bee build inputs with source and safely scanned file hashes for #506.
- Clear stale acceptance files when artifact validation fails.
- Reuse disposable cleanup scopes across affected runtime tests.
- Skip source regex scans when the required literal is absent. Preserve every file check and match.

## How we know

All 1,298 script tests and 658 documentation tests pass. Formatting, spelling, repository checks, and profile tests pass.

Two local Unity runs each pass 1,495 cases with the same expected skip. Existing outcomes remain unchanged after two documented case renames. The suite gains 38 passing cases. Nine emitter regressions fail before the fix and pass afterward.

The source scan preserves all 38 matches across 229 files. All 500 focused tests pass after the cleanup changes. Local timings vary; native CI coverage and timing acceptance are pending.

Independent review reports no required revisions.

## Native upload RCA resolution (commits caca2991, 56143fc3)

The Windows Bee build graph `Playereb61d75c.dag.json` carries Unicode format-control characters. The artifact redactor refused any reviewed text file containing one, so no native perf evidence could upload. Two diagnostics rounds resolved the entire class:

1. A literal format-control character or lone surrogate inside a decoded JSON scalar (category `scalar-format-control`).
2. A structured scalar beginning with a byte-order mark that the offset-0 carve-out preserved, plus escape text that the shadow decoder resolves back into an invisible character (category `unsafe-structured-data`).

Redaction now rewrites literal characters, lone surrogates, JSON u-escape runs (valid surrogate pairs preserved verbatim), and XML numeric entity chains into visible `[cf:xxxx]` markers, counted as `scalar-format-control`. The byte-order-mark carve-out applies only at document level, so a scalar led by one is neutralized too. Encoded-Cf shadows, control-character keys, and structural defects still refuse. Every rewritten output rescans clean, and a second pass is a byte-level no-op.

Two vector rows moved from refusal to exact-byte acceptance and their raw sources now also block immutable sealing. The JS line budget history entry 087 records the reviewed raise to 24060. An adversarial probe then found one more defect in the same branch: an out-of-range numeric entity such as &#x110000; threw RangeError and the blanket catch silently discarded the file's redaction work. Commit 412b881a bounds the parsed code point, out-of-range entities stay verbatim, two safe-evidence rows pin the behavior, and a verifying probe battery (differential across commits plus a 6000-case structured fuzz) reports zero invariant breaks. The budget ceiling is final at 24063. Full npm suite 1298/1298, validate:all, spelling, and the budget gate pass locally.

## CI acceptance at 56143fc3

All 46 check runs pass: 37 success, 8 intentional skips, 1 neutral automated review. The four editor jobs and both perf jobs succeed. The internal perf leg redacted the exact previously rejected file (`Playereb61d75c.dag.json: scalar-format-control`) and uploaded `perf-6000.5.2f1-standalone-internal.zip` (198,665 bytes). Both native artifacts now publish.

Timing against the merged-main baseline (same six native jobs): Unity editor execution 126-146s versus 126-150s baseline (within noise), and the perf evidence redaction phase grew by about 10s per perf job from scanning the newly retained graph and input files. No existing Unity test leg slowed down.

## Remaining work

These changes advance #506, #508, and #509. Full campaign acceptance, native executable identity, and independent immutable restoration remain open in those issues.

---

> [!NOTE]
> **Medium Risk**
> Changes affect perf evidence sealing, artifact redaction (credential/upload safety), and differential replay that drives production global bus override APIs; most messaging logic changes are test-side, but redaction mistakes could leak or over-redact CI artifacts.
> 
> **Overview**
> Extends **differential bus trace replay to generator version 10** with global message-bus override operations (acquire/copy/dispose/replace), lease validation, and replay through `MessageHandler` override APIs. Adds regression coverage for out-of-order disposal, stale copies, generation boundaries, cleanup failures, and primary vs alternate emitter behavior.
> 
> Adds a new **`allocation-subunsub-observations` perf evidence class** that seals and replays raw SubUnsub benchmark CSV/log pairs with strict canonical eight-column validation and independent `-1` unmeasured sentinels for allocation fields.
> 
> For **#506 native perf provenance**, CI now captures Bee graph/input companions and log-referenced `.rsp` files into `native-build-inputs`, then finalizes `retained-manifest.json` with scanned retained hashes during artifact redaction. **Format-control characters** in reviewed text (literals, surrogates, JSON/XML escape chains) are rewritten to visible `[cf:xxxx]` markers so previously blocked graph JSON can upload; stale retained manifests are removed when validation fails.
> 
> Shared **`CleanupScope`** replaces duplicated test teardown patterns; global bus tests gain parameterized generation cases. Docs and JS tests cover the new evidence and redaction behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 412b881a9fb57394cfb8eb9c6afee963b0aca4ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Tests/Runtime/Core/MessageHandlerGlobalBusTests.cs
```diff
@@ -488,38 +492,50 @@ public void ExhaustedGenerationSlotIsBurnedAndStaleCopyCannotEndReplacement()
...
             }
             finally
             {
                 MessageHandler.SetGlobalMessageBus(original);
                 exhaustedScope.Dispose();
                 replacement.Dispose();
-                RestoreBurnedOverrideSlot(exhaustedSlot, originalGeneration);
+                if (generation == -2)
+                {
+                    RestoreBurnedOverrideSlot(exhaustedSlot, originalGeneration);
```
wallstop, line 491: Is it possible to utilize IDisposable mechanisms and avoid try/finally in tests, here and throughout?

wallstop: Implemented in fc3927cf. The existing idempotent CleanupScope is now shared by the runtime tests. All three restoration blocks in this fixture and both logging-restoration blocks in DifferentialBusTraceTests use disposable scopes. Adapter cleanup also uses its existing GlobalBusScope.

Explicit Dispose calls remain where the test checks copied, stale, and out-of-order leases. Callback finalization keeps try/finally where it records behavior after exceptions; this adds no per-emission allocation.

All 500 focused Unity cases pass. Two ordinary runtime runs each preserve all 1,495 passes and the same expected skip. Independent review confirms cleanup order, exception behavior, test inputs, and assertions are unchanged.

